PR #110, opened 2020-11-16
rc-4.0.0

ivanklee86:
Release Candidate for 4.0.0.

```
**General**
* (Major) Stop using the `default_value` argument in the `is_enabled()` method (as it can cause counter-intuitive behavior) and add deprecation warning.  This argument will be removed in the next major version upgrade!  
    * We recommend using the `fallback_function` argument instead.  If you need a blanket True in case of an exception, you can pass in a lambda like: `lambda x, y: True`. 
* (Minor) Add better logging for API errors.
* (Minor) Bump requests version.
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.4.2
+current_version = 4.0.0
 commit = True
 tag = True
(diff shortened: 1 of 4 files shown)


ivarconr (approved):
I am a bit unsure if this should be a 4x release.

PR #127, opened 2021-02-04
Fix issue with updating in-memory variants in the loader

simenaasland:
# Description

When the new features are downloaded, the loader first updates the file cache before updating
the in-memory cache.

This PR attempts to fix an issue where the loader did not properly update the variants part of the cache.
This caused the variants to be stale until the next cache initialisation on client restart.

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/UnleashClient/loader.py
+++ b/UnleashClient/loader.py
@@ -95,7 +95,7 @@ def load_features(cache: FileCache,
                 feature_for_update.strategies = parsed_strategies
 
             if 'variants' in parsed_features[feature]:
-                feature_for_update.variants = Variants(
+                feature_for_update.variations = Variants(
                     parsed_features[feature]['variants'],
                     parsed_features[feature]['name']
                 )
--- a/tests/unit_tests/test_loader.py
+++ b/tests/unit_tests/test_loader.py
@@ -8,9 +8,6 @@ from tests.utilities.mocks import MOCK_ALL_FEATURES
 from tests.utilities.testing_constants import DEFAULT_STRATEGY_MAPPING
 from tests.utilities.decorators import cache_full, cache_custom  # noqa: F401
 
-MOCK_UPDATED = copy.deepcopy(MOCK_ALL_FEATURES)
-MOCK_UPDATED["features"][4]["strategies"][0]["parameters"]["percentage"] = 60
-
 
 def test_loader_initialization(cache_full):  # noqa: F811
     # Set up variables
@@ -41,7 +38,7 @@ def test_loader_initialization(cache_full):  # noqa: F811
             assert strategy.variants
 
 
-def test_loader_refresh(cache_full):  # noqa: F811
+def test_loader_refresh_strategies(cache_full):  # noqa: F811
     # Set up variables
     in_memory_features = {}
     temp_cache = cache_full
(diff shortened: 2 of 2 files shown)


ivarconr:
Awesome, thanks! 
Funny that these bugs are allowed to live for so long! 

All in all a good catch. I would however prefer that the internal of Feature just used `variants` instead of `variations` (https://github.com/Unleash/unleash-client-python/blob/master/UnleashClient/features/Feature.py#L24). This would have kept the naming of things close the the protocol, and we would probably not have fooled ourselves.

ivanklee86:
@simenaasland Ack, good catch!  Thanks for the PR!

@ivarconr I can certainly line up the names better, though I'd prefer to do that in a separate PR.  I'll get to it in the next few days.

I'll do a bugfix release later today/early tomorrow with this fix. :)

simenaasland:
@ivarconr @ivanklee86 Great that the fix is appreciated, happy to help! It makes sense to use `variants` in `Feature`. We considered changing it that way, and can amend the PR now. Totally up to you.

@ivanklee86 Thanks a lot for making a quick bugfix release! Let us know if you want us to do any renaming here, or if you want to take care of that. :)

ivanklee86:
If you've got time, I'd be more than happy to include the renaming in this PR. :) If not, I'll try and tackle it next time I get some coding time. ^^

simenaasland:
> If you've got time, I'd be more than happy to include the renaming in this PR. :) If not, I'll try and tackle it next time I get some coding time. ^^

I have time to fix it. However, having thought about it a bit more, I'm wondering if there is any risk involved in changing the attribute name without keeping the old one. If external users for some reason looked up the old attribute on Feature, it could be a problem. I assume this is unlikely, but I wanted to confirm with you before making a potential breaking change. :)

ivanklee86:
Hmm, fair point. 

I'd propose:
* Making a bugfix on the current semver (3.6.1) to fix the problem for everyone.
* Rename the variable in the next major version bump (4.x.x) - there's another deprecation queued up.

It's not the best reason to do a major version bump, but I think it'd be OK to clearly communicate to users that there's backwards incompatible changes.

Does that make sense?  If so, I think we can merge this PR (to fix the issue)  and I can create a v4.0.0 branch for the next major release.

simenaasland:
>>> Does that make sense? If so, I think we can merge this PR (to fix the issue) and I can create a v4.0.0 branch for the next major release.

That sounds like a good plan to me. :)

ivanklee86:
I've released v3.6.1 with the fix and created the v4.0.0 branch for any backwards incompatible changes. ^^

ivarconr:
Sounds like a plan! 

fyi: changing `variations` on `Feature` class will only be a breaking if this class is ever returned out of the SDK.

simenaasland:
> I've released v3.6.1 with the fix and created the v4.0.0 branch for any backwards incompatible changes. ^^

Thank you for the super quick release, much appreciated!

> fyi: changing variations on Feature class will only be a breaking if this class is ever returned out of the SDK.

That is true, I don't see it as a high risk, just wanted to be extra careful. :)

PR #134, opened 2021-03-03
client: don't use mutable defaults

aviau:
This does not do what you think.

Defaulting to a dict will default the **same** dict for every call of that function.

Read about mutable defaults, a common python mistake:
 - https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Creating the dict inside the function and not inside the method definition ensures a new empty dict every time.

This is especially important for `is_enabled`'s  context, since the library modifies it.


ivanklee86:
Thanks for the PR!  Yes, this makes sense to fix and the linting issue is unrelated (and hopefully will be fixed in a separate PR).

PR #135, opened 2021-03-04
Remove dependancies version pinning

isra17:
# Description

Removes the dependancies version constraint.

This project is intended to be used as a library, there's no reason why it should et specific version for its dependancies. This cause many issue to application using this module (Version conflict, cannot upgrade dependancies).

Unless this project has a really good reason to require a specific release, this should remove any constraint althogether.
This also save this project from making a new release every time one of its dependancies has a new release.

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests


wbolster (approved):

wbolster:
it would be great if this could be merged and released.

the current approach uses exact pinning, and that is ​quite bad behaviour for a library, since installing it prevents applications from upgrading dependencies.

specifically, newer `mmh3` versions have upstream `manylinux` wheels, which makes installing this library *a lot* simpler since no compilation step is required.

see https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files for more info, as well as most other python libs that do not use exact pinning. :upside_down_face:

wbolster:
the ci build failure is unrelated; i have opened #137 to fix those.

ivanklee86:
I think it makes sense to unpin the dependencies, though I'm thinking of doing so in a major version out of an abundance of caution since `mmh3` drops support for Python 3.5.  Thoughts on how you'd like to see it released?

cc @ivarconr

isra17:
I don't see any issue with major release!

wbolster:
3.5 is end of life. but even ignoring that, removing the pins allows users to keep using an older mmh3 version, so the situation does not get worse in any way.

the new mmh3 version comes with binary wheels, making it much easier to install mmh3, and hence also unleashclient, which is obviously a good thing!

wbolster:
@ivanklee86 fyi, you may want to get rid of the versions in `requirements.txt` as well 💡

ivanklee86:
Yup, I'll include that in a separate cleanup PR. :)

ivarconr:
> I think it makes sense to unpin the dependencies, though I'm thinking of doing so in a major version out of an abundance of caution since `mmh3` drops support for Python 3.5. Thoughts on how you'd like to see it released?

This should absolutely be a 4x release. I am not sure how widely used 3.5 is anymore?

isra17:
> @ivanklee86 fyi, you may want to get rid of the versions in `requirements.txt` as well 💡

I don't think this is right. The point of requirements.txt is to create a reproducible environment for development and testing, therefor pinning dependancies make sense there.

wbolster:
pinning in `requirements.txt` is totally right… for applications.

but this project is a library, and it does not make sense to always the tests against an environment that is guaranteed to be partially outdated very soon.

right now some of the (test) dependencies are pinned, while many others were not. that is not a good situation to be in in any case. #139 by @ivanklee86 addresses this by removing all of them. (which would  also be my personal preference.)

as a result, incompatibilities because of ecosystem upgrades will pop up whenever CI runs. the same goes for updated linting tools that suddenly complain about new things, see e.g. #137.

the alternative is to pin everything, and let dependabot help keep it up-to-date. the result is much more chore work and many more (potentially expensive) CI builds. i personally don't think this is worth it. libraries like this don't get autodeployed from the `master` branch (or a pull request) anyway.

and if there is a sudden incompatibility that cannot be addressed right away, it's easy to add a temporary `foo<=x.y` pin, with `x.y` denoting the highest know working version of `foo`, and address it later.

PR #190, opened 2022-01-25
because of unleash proxy server, added default for strategies key and copied toggles key into features

salva-imm:
# Description

i discovered unleash proxy server's response is different from gitlab API
its like this:
{'toggles': [{'name': 'some-feature', 'enabled': True, 'variant': {'name': 'disabled', 'enabled': False}}]}

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/UnleashClient/loader.py
+++ b/UnleashClient/loader.py
@@ -73,7 +73,7 @@ def load_features(cache: FileCache,
         feature_provisioning = cache[FEATURES_URL]
 
         if not feature_provisioning.get("features"):
-            feature_provisioning['features'] = feature_provisioning.get('toggles')
+            feature_provisioning['features'] = feature_provisioning.get('toggles', {})
 
         # Parse provisioning
         parsed_features = {}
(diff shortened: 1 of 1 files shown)


ivarconr:
Hi there. The proxy format is different on purpose, and you are not supposed to connect a server side SDK directly against the proxy endpoint. The main reason for introducing the unleash proxy was to serve single page and native applications and the proxy will evaluate feature toggles for a certain context. This is similar to how a server side SDK operates. 

It's nothing stopping you from using the proxy from Python, but that would be a simple HTTP call, and you would probably not need an SDK to do that. 

You can read more about the proxy here:
https://docs.getunleash.io/sdks/unleash-proxy

cwirz:
@ivarconr could you please explain why one is not supposed to use the proxy for this on the server side? its not documented that this is not possible anywhere or did i miss something?

cwirz:
@ivarconr thanks for the response. much appreciated.

Issue #191, opened 2022-02-03
feat: Add support for new constraint operators


ivarconr:
I have written a summary In this issue: https://github.com/Unleash/unleash/issues/856#issuecomment-1027910183

Client-specifications: https://github.com/Unleash/client-specification/pull/25

Node SDK reference implementation: https://github.com/Unleash/unleash-client-node/pull/289

**PS! we are still working out the UI and validating the actual protocol, but we are at > 93% certainty.**

ivanklee86:
Tasks:
- [ ] Add `currentTime` to default Unleash context.
- [ ] Modify [Constraint initialization](https://github.com/Unleash/unleash-client-python/blob/main/UnleashClient/constraints/Constraint.py#L6) for new values (`caseInsensitive` -> `case_insensitive`, `inverted` 'value')
- [ ] Implement new conditionals.
- [ ] Add unit tests to Constraint object.
- [ ] Add spec tests.

Resources:

ivarconr:
can we close this issue? 
https://github.com/Unleash/unleash-client-python/blob/main/CHANGELOG.md#v510

ivanklee86:
Totally forgot about this issue.  It can indeed be closed!

PR #192, opened 2022-02-04
Add new constraint operators

ivanklee86:
# Description

Supports new operators outlined in #191 .

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -1,8 +1,7 @@
 # pylint: disable=invalid-name
 import warnings
 from datetime import datetime, timezone
-from typing import Dict, Callable, Any, Optional
-import copy
+from typing import Callable, Optional
 from fcache.cache import FileCache
 from apscheduler.job import Job
 from apscheduler.schedulers.background import BackgroundScheduler
@@ -227,7 +226,13 @@ class UnleashClient:
         :return: Feature flag result
         """
         context = context or {}
+
+        # Update context with static values
         context.update(self.unleash_static_context)
+
+        # Update context with optional values
+        if 'currentTime' not in context.keys():
+            context.update({'currentTime': datetime.now()})
 
         if self.is_initialized:
             try:
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -234,10 +234,6 @@ class UnleashClient:
         # Update context with static values
         context.update(self.unleash_static_context)
 
-        # Update context with optional values
-        if 'currentTime' not in context.keys():
-            context.update({'currentTime': datetime.now()})
-
         if self.is_initialized:
             try:
                 return self.features[feature_name].is_enabled(context)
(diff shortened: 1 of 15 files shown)


UnleashClient/__init__.py
```diff
@@ -227,8 +226,14 @@ def is_enabled(self,
         :return: Feature flag result
         """
         context = context or {}
+
+        # Update context with static values
         context.update(self.unleash_static_context)
 
+        # Update context with optional values
+        if 'currentTime' not in context.keys():
+            context.update({'currentTime': datetime.now()})
```
ivarconr, line 235: I might need to steal this to the Node.SDK. 

The only reason I kept this in the operator itself is that it does give an additional cost for every `isEnabled` call to create currentTime instance, even if it is mostly not needed (which will be 99% of the cases). On the other hand, datetime.now() probably has an insignificant cost?

UnleashClient/constraints/Constraint.py
```diff
@@ -1,17 +1,150 @@
...
 
         :param constraint_dict: From the strategy document.
         """
-        self.context_name = constraint_dict['contextName']
-        self.operator = constraint_dict['operator']
-        self.values = constraint_dict['values']
+        self.context_name: str = constraint_dict['contextName']
+        self.operator: ConstraintOperators = ConstraintOperators(constraint_dict['operator'].upper())
+        self.values = constraint_dict['values'] if 'values' in constraint_dict.keys() else []
+        self.value = constraint_dict['value'] if 'value' in constraint_dict.keys() else []
```
ivarconr, line 47: else `[]'  seems a bit weird for a string?

ivarconr: Maybe `None` is better?

ivanklee86: Whoops, that's an embarassing C/P error.  Def. should be None!

UnleashClient/__init__.py
ivanklee86: tldr after a few rabbit holes: Yes for modern CPUs.  

I was mainly putting it there since it seemed that it might be handy for other uses (on reflection though, not 100% sure what those could be).  I'd be OK with moving this into the `Constraint` check.  

Do you think you'd use `currentTime` in other features?

ivarconr: not sure yet. I do not see a big issue moving it out to the context creation later if we find use for it. I would prefer us to have it in the constraint for now.

ivanklee86: I've moved this down into the `Constriant.apply()` function: https://github.com/Unleash/unleash-client-python/pull/192/files#diff-5a34048a72d9dbcc1cf64fe80a13119b0b40ee924c5cfc90a0352c30e7036ac4R162

Issue #194, opened 2022-02-15
Make cache implementation plugable


Meemaw:
**Is your feature request related to a problem? Please describe.**
We're seeing some issues with the `fcache` implementation. Its not multi process friendly, has race conditions, and is not maintained.

**Describe the solution you'd like**
a) Switch to a better fcache solution, e.g. https://github.com/grantjenks/python-diskcache
b) Make the cache pluggable, so users can switch the implementation

**Describe alternatives you've considered**
N/A

damjankuznar:
:heavy_plus_sign:

povilasb:
Isn't it already: https://github.com/Unleash/unleash-client-python/blob/aafb389267417b70f20613c42d455e0a6847716f/UnleashClient/__init__.py#L55 ? :)

povilasb:
I guess this can be closed since https://github.com/Unleash/unleash-client-python/pull/193 was merged?
@damjankuznar FYI, I've been using Unleash with `diskcache` - works well so far.

Issue #105, opened 2020-09-15
In-memory features are not updated when initialized from a Celery worker


kmkr:
**Describe the bug**

We use unleash-client-python in our code in both gunicorn and Celery environments. The `fetch_and_load_features` method runs in the background in both cases. However, updates are only propagated to the client in gunicorn environments. Clients initialized in Celery environments don't receive updates to it's _memory_ instance (`self.features`). Updates _are_ propagated to the cache, so I can workaround the problem by running `load_features` in the `is_enabled` function:

```python
  def is_enabled(self,
                   feature_name: str,
                   context: dict = {},
                   default_value: bool = False,
                   fallback_function: Callable = None) -> bool:
        # Docstring removed for brevity
        context.update(self.unleash_static_context)

        # Workaround the problem by loading features from the cache:
        load_features(self.cache, self.features, self.strategy_mapping)

        if self.is_initialized:
            try:
                return self.features[feature_name].is_enabled(context, default_value)
       # more code removed for brevity
```

The problem is perhaps related to how the `apsscheduler.BackgroundScheduler` works/not works together with Celery, but I'm not sure. 

Do you have any suggestions to solve this problem--without manually loading the features from cache?

**To Reproduce**

1. Initiate the Unleash client in a Celery worker
2. Wait for the first `fetch_and_load_features` to complete.
3. Update a toggle in the Unleash interface, for example from enabled to disabled.
4. Wait for the background worker to fetch and load features.
5. Call `is_enabled` on the feature. The feature's state is not updated but should be.

ivanklee86:
Hello!

I, unfortunately, haven't really given unleash-client-python a shot in Celery.  I'll give it a try but may take a lil' bit as work is being work-ey.

I had a few general questions:
* Are you running [Celery in Flask](https://flask.palletsprojects.com/en/1.1.x/patterns/celery/) (or the app-factory equivalent)?  
* If you set the log level to DEBUG you should see some log statements from apscheduler (example below).  By any chance is anything showing up in the celery task?

```
2020-09-15 13:01:32,216 - apscheduler.scheduler - INFO - Scheduler started
2020-09-15 13:01:32,216 - apscheduler.scheduler - DEBUG - Looking for jobs to run
2020-09-15 13:01:32,216 - apscheduler.scheduler - DEBUG - No jobs; waiting until a job is added
2020-09-15 13:01:32,217 - apscheduler.scheduler - INFO - Added job "fetch_and_load_features" to job store "default"
[...]

PR #205, opened 2022-04-22
2 modes for running Unleash integration

povilasb:
# Description

Usually Unleash Python SDK would do 2 things periodically:

1. fetch feature toggles from the Unleash server,
2. load them into the local [Cache](https://github.com/Unleash/unleash-client-python/blob/4c7bc98e7919f5f3ae91ca1c732337be61b2d17b/UnleashClient/cache.py#L7).

Then SDK uses the toggles from the cache.

However, when a service (or application) is using multiprocessing for concurrency, such model creates higher loads on Unleash than ideal:

1. now each separate process has it's own cache and is fetching the toggles. This is somewhat touched at https://github.com/Unleash/unleash-client-python/issues/105
2. Instead, we'd like one process to fetch the toggles and other process load them.

To cater such concurrency models this PR allows to run Unleash SDK in 2 modes:
1. The regular way: fetch + load: `UnleashClient().initialize_client()`
2. Only load the toggles from the cache, but do not fetch them from the server: `UnleashClient().initialize_client(fetch_toggles=False)`

Obviously, this is backwards compatible.

Now this by itself doesn't allow to use this SDK in multiprocess environment, however in conjuction with [diskcache](https://pypi.org/project/diskcache/) as mentioned in [this issue](https://github.com/Unleash/unleash-client-python/issues/194) seems to operate fine.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

The changes are quite trivial, hence only manual testing.

- [ ] Unit tests
- [ ] Spec Tests
- [X] Manual

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -10,6 +10,7 @@ from UnleashClient.periodic_tasks import fetch_and_load_features, aggregate_and_
 from UnleashClient.strategies import ApplicationHostname, Default, GradualRolloutRandom, \
     GradualRolloutSessionId, GradualRolloutUserId, UserWithId, RemoteAddress, FlexibleRollout
 from UnleashClient.constants import METRIC_LAST_SENT_TIME, DISABLED_VARIATION, ETAG
+from UnleashClient.loader import load_features
 from .utils import LOGGER
 from .deprecation_warnings import strategy_v2xx_deprecation_check
 from .cache import BaseCache, FileCache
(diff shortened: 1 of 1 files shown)


ivanklee86:
Thanks for the PR!  This makes sense, is backwards compatible, and clear to the users what they're getting into. =)

ivanklee86 (approved):
Thanks for the PR! This makes sense, is backwards compatible, and clear to the users what they're getting into. =)

I have another feature in the pipeline but I'll try and push a new version containing this early next week.

PR #210, opened 2022-05-31
docs: add a section on getting the client working with celery

sighphyre:
Adds documentation on how to get this project working with Celery

### DIFF
--- a/docs/celery.rst
+++ b/docs/celery.rst
@@ -1,0 +1,16 @@
+****************************************
+Running with Celery
+****************************************
+
+To use this client with Celery, call the `initialize_client` method in the `worker_process_init` event. If you don't initialize the client inside the `worker_process_init` event, the client will not poll for updates. As such, the client will not update its feature toggles, even when changes are made server side.
+
+.. code-block:: python
+
+    client = UnleashClient(
+        url="http://localhost:4242/api/",
+        app_name="test-dev",
+        custom_headers={'Authorization': '*:development.bb09e81624d5ad67b2ac29bd0b0fdc35ccbac884e63cfd20c6fefc49'})
+
+    @worker_process_init.connect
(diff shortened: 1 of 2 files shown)


docs/celery.rst
```diff
@@ -0,0 +1,15 @@
+****************************************
+Running with Celery
+****************************************
+
+The client is also compatible with Celery, however, the initialize_client method needs to be called in the worker_process_init event, otherwise toggles will not update correctly when changes on the server.
```
thomasheartman, line 5: Nice, but maybe we could rephrase it to something like this?

```suggestion
To use this client with Celery, call the `initialize_client` method in the `worker_process_init` event. If you don't initialize the client inside the `worker_process_init` event, toggles will not update correctly when changes occur on the server.
```

What does "update correctly" mean here? Will they not update at all or just at weird times?

thomasheartman (changes requested):
Good one! 🙌🏼  Should be great to get this in 😄  I've added a small rewording suggestion and have a little discussion point before we do, though.

docs/celery.rst
sighphyre: Oh that's much better, thanks for the suggestion. Not "update correctly" should probably be "not update at all" 

What about this (just removing the "correctly"):

To use this client with Celery, call the `initialize_client` method in the `worker_process_init` event. If you don't initialize the client inside the `worker_process_init` event, toggles will not update when changes occur on the server.

thomasheartman: Yeah, that's great! Might even wanna **bold** or _italicize_ the "not" in "will not update". But maybe it can be reworded a bit. Something like:

> If you don't initialize the client inside the `worker_process_init` event, the client will not poll for updates. As such, the client will not update its feature toggles even when changes are made server side.

PR #212, opened 2022-06-21
Feature/global segments

sighphyre:
This adds support for reusable global segments, reference implementation in node can be viewed here: https://github.com/Unleash/unleash-client-node/pull/329

This also includes the header for turning on segments

### DIFF
--- a/UnleashClient/variants/Variants.py
+++ b/UnleashClient/variants/Variants.py
@@ -79,7 +79,7 @@ class Variants:
             if override_variant:
                 return self._format_variation(override_variant)
 
-            total_weight = sum([x['weight'] for x in self.variants])
+            total_weight = sum(x['weight'] for x in self.variants)
             if total_weight <= 0:
                 return fallback_variant
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,9 @@ addopts= """
 """
 log_file_level="INFO"
 
+[tool.pylint]
+load-plugins="pylint.extensions.no_self_use"
+
 [tool.pylint.messages_control]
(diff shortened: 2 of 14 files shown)


UnleashClient/variants/Variants.py
```diff
@@ -79,7 +79,7 @@ def get_variant(self, context: dict) -> dict:
             if override_variant:
                 return self._format_variation(override_variant)
 
-            total_weight = sum([x['weight'] for x in self.variants])
+            total_weight = sum(x['weight'] for x in self.variants)
```
sighphyre, line 82: Swaps a list comprehension for a generator because Pylint doesn't like this

pyproject.toml
```diff
@@ -15,6 +15,9 @@ addopts= """
 """
 log_file_level="INFO"
 
+[tool.pylint]
+load-plugins="pylint.extensions.no_self_use"
```
sighphyre, line 19: Pylint got updated, the no_self_use option is now opt in so this turns that back on

chriswk (approved):
The logic seems sound, but have to admit my Python is quite rusty, the implementation seems close to how we do it in Node and Java

chriswk:
[Segmentation SDK support: Python](https://trello.com/c/vDW3UZxF/1285-segmentation-sdk-support-python)

ivanklee86:
Could I get a link to a document outlining Segmentation (or maybe the Trello ticket)?  Just wanna 👀 to make sure I understand what's supposed to be happening. 😄 

(Sorry, busy this week 'cause going on PTO next week.  Should offer more feedback

PR #231, opened 2023-01-05
Feat/block multiple constructions

sighphyre:
This introduces a new instance checking concept into the SDK, this is a result of regular misconfigurations across the board with the SDKs, which result in multiple polling instances hammering the server. By default the instance checking will raise a warning if the constructor is called more than once in a process with the same configuration. This also allows the consumer to change this to either not log the warning at all, in rare cases that multiple clients may be correct, or to raise the warning to an error and prevent multiple instantiations at all.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,8 @@
 
 [tox]
 envlist = py37,py38,py39,py310
-allowlist_externals = ./get_spec.sh
+allowlist_externals =
+    ./get_spec.sh
 
 [testenv]
 deps = -rrequirements.txt
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 
 [tox]
 envlist = py37,py38,py39,py310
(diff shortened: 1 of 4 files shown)


sighphyre:
Build seems to be broken on this. Doesn't seem to be related to these changes (same thing happens on a new branch from main and the stuff that fails isn't related to this), but I am looking into it

gastonfournier (approved):
Looks good!

tox.ini
```diff
@@ -4,9 +4,10 @@ envlist = py37,py38,py39,py310
 
 [testenv]
 deps = -rrequirements.txt
+allowlist_externals = sh
 commands =
     mypy UnleashClient --install-types --non-interactive
-    ./get-spec.sh
+    sh get-spec.sh
```
sighphyre, line 10: Fixes an issue with an update to Tox latest version

PR #237, opened 2023-02-24
feat: Extend Feature object with new attributes

akalex:
# Description

Currently, there is no way to get answers to the question - **Whether impression data is enabled?**

The good news is that this information is provided so we can store it and use it if needed.

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:

### DIFF
--- a/UnleashClient/loader.py
+++ b/UnleashClient/loader.py
@@ -59,11 +59,13 @@ def _create_feature(provisioning: dict,
     else:
         variant = None
 
-    return Feature(name=provisioning["name"],
-                   enabled=provisioning["enabled"],
-                   strategies=parsed_strategies,
-                   variants=variant
-                   )
+    return Feature(
+        name=provisioning["name"],
+        enabled=provisioning["enabled"],
+        strategies=parsed_strategies,
+        variants=variant,
+        impression_data=provisioning.get("impressionData", False),
+    )
 
 
 def load_features(cache: BaseCache,
@@ -118,6 +120,8 @@ def load_features(cache: BaseCache,
                 parsed_features[feature]['variants'],
                 parsed_features[feature]['name']
             )
+
+        feature_for_update.impression_data = parsed_features[feature].get("impressionData", False)
 
     # Handle creation or deletions
     new_features = list(set(feature_names) - set(feature_toggles.keys()))
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -1,4 +1,6 @@
 # pylint: disable=invalid-name
+from typing import Optional
+
 from UnleashClient.variants import Variants
 from UnleashClient.utils import LOGGER
 from UnleashClient.constants import DISABLED_VARIATION
@@ -10,19 +12,24 @@ class Feature:
                  name: str,
                  enabled: bool,
                  strategies: list,
-                 variants: Variants = None) -> None:
+                 variants: Optional[Variants] = None,
+                 impression_data: bool = False) -> None:
         """
         A representation of a feature object
 
         :param name: Name of the feature.
         :param enabled: Whether feature is enabled.
         :param strategies: List of sub-classed Strategy objects representing feature strategies.
+        :param impression_data: Whether impression data is enabled.
         """
         # Experiment information
         self.name = name
         self.enabled = enabled
         self.strategies = strategies
         self.variants = variants
+
+        # Additional information
+        self.impression_data = impression_data
 
         # Stats tracking
         self.yes_count = 0
--- a/tests/unit_tests/test_features.py
+++ b/tests/unit_tests/test_features.py
@@ -32,6 +32,7 @@ def test_create_feature_true(test_feature):
 
     my_feature.reset_stats()
     assert my_feature.yes_count == 0
+    assert not my_feature.impression_data
 
 
 def test_create_feature_false(test_feature):
--- a/tests/unit_tests/test_loader.py
+++ b/tests/unit_tests/test_loader.py
@@ -36,6 +36,8 @@ def test_loader_initialization(cache_full):  # noqa: F811
         if isinstance(strategy, Variants):
             assert strategy.variants
 
+        assert feature.impression_data is False
+
 
 def test_loader_refresh_strategies(cache_full):  # noqa: F811
     # Set up variables
--- a/tests/utilities/mocks/mock_all_features.py
+++ b/tests/utilities/mocks/mock_all_features.py
@@ -1,214 +1,223 @@
-MOCK_ALL_FEATURES = \
-{
-  "version": 1,
-  "features": [
-    {
-      "name": "ApplicationHostname",
-      "description": "Application Hostname strategy",
-      "enabled": True,
-      "strategies": [
+MOCK_ALL_FEATURES = {
+    "version": 1,
+    "features": [
         {
-          "name": "applicationHostname",
-          "parameters": {
-            "hostNames": "iMacPro.local,test1,test2"
-          }
-        }
-      ],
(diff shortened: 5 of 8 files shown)


tests/specification_tests/test_client_specs.py
```diff
@@ -29,6 +29,9 @@ def load_specs():
 def iter_spec():
     for spec in load_specs():
         name, state, tests, variant_tests = load_spec(spec)
+        state["features"] = [
+            {"impressionData": False, "stale": False, **feature} for feature in state["features"]
+        ]
```
akalex, line 34: I should be able to remove these lines once https://github.com/Unleash/client-specification/pull/51 has merged

UnleashClient/loader.py
```diff
@@ -62,7 +62,9 @@ def _create_feature(provisioning: dict,
     return Feature(name=provisioning["name"],
                    enabled=provisioning["enabled"],
                    strategies=parsed_strategies,
-                   variants=variant
+                   variants=variant,
+                   impression_data=provisioning["impressionData"],
+                   stale=provisioning["stale"]
```
sighphyre, line 67: Looks like these will fail to load if these properties aren't present, which I'd rather avoid. `impression_data` won't be there prior to Unleash 4.7 and I'm not certain we can guarantee either of them in the bootstrap files. Can we swap this out for something more tolerant to these properties being absent like `provisioning.get("impressionData")`?

tests/specification_tests/test_client_specs.py
sighphyre: Hopefully this won't be necessary if we address the previous comment

sighphyre (commented):
Two thoughts here:
1) This seems to add the property for impression data but it doesn't actually add the functionality. Was that intended?

2) `stale` usually isn't actually used meaningfully by the SDKs, it seems like you're wanting to get your hands on this property within the SDK so I'm super curious, what's your usecase here? I'm not against exposing them but usually when I see this kind of pull request, the SDK is being used for something we didn't design them for

akalex:
> stale usually isn't actually used meaningfully by the SDKs, it seems like you're wanting to get your hands on this property within the SDK so I'm super curious, what's your usecase here? I'm not against exposing them but usually when I see this kind of pull request, the SDK is being used for something we didn't design them for

The indention to open this PR mainly was issued by the absence of possibility in Python SDK to get impression data and have no clue when it will be added. So, I learned that extra information is available as part of the data pulled by the provisioning loader, but the object `Feature` does not preserve that information. 
Maybe I use SDK improperly, and you can advise what would be a better approach to get to know about Impression Data. Thank you in advance.

UnleashClient/loader.py
akalex: Sure, I can make it fail-safe. I didn't know such details, so I thought that should be safe.

sighphyre:
> > This seems to add the property for impression data but it doesn't actually add the functionality. Was that intended?
> 
> > stale usually isn't actually used meaningfully by the SDKs, it seems like you're wanting to get your hands on this property within the SDK so I'm super curious, what's your usecase here? I'm not against exposing them but usually when I see this kind of pull request, the SDK is being used for something we didn't design them for
> 
> The intention to open this PR mainly was issued by the absence of possibility in Python SDK to get impression data and have no clue when it will be added. So, I learned that extra information is available as part of the data pulled by the provisioning loader, but the object `Feature` does not preserve that information. Maybe I misuse SDK, and you can advise what would be a better approach to get to know about Impression Data. I really appreciate any help you can provide.

Right, this sounds like what you're doing here is starting the work/giving us a gentle nudge on implementing impression data in this SDK? That makes a lot of sense, that's absolutely not a misuse of this SDK. But this is a bigger feature than just implementing the properties that come back from the server. It's probably worth going going over the PRs that add this feature to the dotnet SDK (https://github.com/Unleash/unleash-client-dotnet/pull/113) and the node SDK (https://github.com/Unleash/unleash-client-node/pull/336) if you're interested in the technical details here and are looking to implement this here.

What probably needs to be done here is exposing the impression data property (done already), adding an event system or using Python built ins to expose a stream of events and then injecting the events into that system when the toggle is checked.

It sounds like this is also a request for us to implement this feature, so I've opened a ticket here for it: https://github.com/Unleash/unleash-client-python/issues/238 so that if this is more work than you had wanted to get involved we can track it internally.

akalex:
> Right, this sounds like what you're doing here is starting the work/giving us a gentle nudge on implementing impression data in this SDK? That makes a lot of sense, that's absolutely not a misuse of this SDK. But this is a bigger feature than just implementing the properties that come back from the server. It's probably worth going going over the PRs that add this feature to the dotnet SDK (https://github.com/Unleash/unleash-client-dotnet/pull/113) and the node SDK (https://github.com/Unleash/unleash-client-node/pull/336) if you're interested in the technical details here and are looking to implement this here.

> What probably needs to be done here is exposing the impression data property (done already), adding an event system or using Python built ins to expose a stream of events and then injecting the events into that system when the toggle is checked.

> It sounds like this is also a request for us to implement this feature, so I've opened a ticket here for it: https://github.com/Unleash/unleash-client-python/issues/238 so that if this is more work than you had wanted to get involved we can track it internally.

You are correct that would be absolutely great if you could prioritise some work around Impression Data, and thanks for opening a ticket.

On the other hand, I still do not see a strong reason why the `Feature` object should not be extended with the properties that the object already has. Such changes neither break interface nor go against the rules but rather provides fair compromise and room for maneuvers.

What are your thoughts on that?

sighphyre:
@akalex I'm very open to including the inclusion of of the impression data property before we've implemented the feature itself but I'm still not sure what including the stale property would add. Do you see a use case here that I'm missing?

akalex:
> @akalex I'm very open to including the inclusion of of the impression data property before we've implemented the feature itself but I'm still not sure what including the stale property would add. Do you see a use case here that I'm missing?

@sighphyre that's a great question, I don't have a good example from the top of my head. Let me get rid of the `stale` property.

UnleashClient/loader.py
akalex: this is done.

akalex:
> > @akalex I'm very open to including the inclusion of of the impression data property before we've implemented the feature itself but I'm still not sure what including the stale property would add. Do you see a use case here that I'm missing?
> 
> @sighphyre that's a great question, I don't have a good example from the top of my head. Let me get rid of the `stale` property.

@sighphyre this is done. could you take a look at it once you have a chance

sighphyre:
@akalex Thanks so much for doing the changes. This looks good to me. To keep you in the loop, I've raised the impression data feature with the team internally to look into next week, we'll probably just build on top of this PR

sighphyre (approved):
LGTM

akalex:
> @akalex Thanks so much for doing the changes. This looks good to me. To keep you in the loop, I've raised the impression data feature with the team internally to look into next week, we'll probably just build on top of the work in this PR

@sighphyre thank you so much, I appreciate your help in pushing this one further. One last question if you don't mind, when these changes optimistically could be released?

sighphyre:
Hey @akalex, probably when we finish off the feature, can't make promises until the team plans this but I will raise it for next week, I'm not sure there's a point in doing a release until this implements the feature fully, right now it's just a data property

akalex:
> Hey @akalex, probably when we finish off the feature, can't make promises until the team plans this but I will raise it for next week, I'm not sure there's a point in doing a release until this implements the feature fully, right now it's just a data property

@sighphyre It's data property, but even without the Impression data feature, it allows us to build and use custom solutions like in my case. So I either have to fork and support that I don't really want to or you can release it as a new version.

PR #249, opened 2023-04-24
Fix warning scheduler_executor without scheduler

nhubanov:
# Description

Removes warning message when Unleash client is initialised without scheduler and scheduler_executor.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Unit tests
- [ ] Spec Tests

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -139,7 +139,7 @@ class UnleashClient:
                 "If using a custom scheduler, you must specify a executor."
             )
         else:
-            if not scheduler:
+            if not scheduler and scheduler_executor:
                 LOGGER.warning(
                     "scheduler_executor should only be used with a custom scheduler."
                 )


tests/unit_tests/test_client.py
```diff
@@ -682,7 +682,7 @@ def test_multiple_instances_no_warnings_or_errors_with_different_client_configs(
     UnleashClient(
         URL, "some-probably-unique-but-different-app-name", refresh_interval="60"
     )
-    assert not all(
+    assert not any(
```
sighphyre, line 685: I think the change looks good but I'm not sure why these tests needed to change. Makes me a little nervous when we have to change unrelated tests and there's no new tests to cover the behaviour

nhubanov: These tests are implicitly related to the change. 

Before it every time you create an UnleashClient without `scheduler` you get a warning log record. `not all(...)` was working, because the `msg` property of the record do not containe "Multiple instances has been disabled".  With this change there is no longer warning log records and `not all([])` becomes `False`. Logically the assertion should assert there isn't any log message about "Multiple instances has been disabled".

sighphyre: Sure, makes sense! LGTM

PR #253, opened 2023-05-24
feat: add variant metrics

thomasheartman:
## Description

This change adds most variant metrics functionality to the Python SDK.
Variant metrics should be counted whenever get_variant is called:

- if a feature doesn't have variants, register the disabled variant
- if a feature is disabled, register the disabled variant
- otherwise, register the variant that was returned

There is one piece of functionality that isn't included in this PR:

- if a feature doesn't exist, it should still register the disabled variant.

As far as I can tell, this SDK doesn't count metrics for toggles that don't exist, so there was no obvious way to add variant metrics for these features. Both of those cases will be implemented in a follow-up PR.

### Serialization of emtpy dicts

When serializing metrics, I have done the simple thing and added the variant counts dict, regardless of whether it is empty or not. It might be preferable to not include that key at all if the feature doesn't have any variant metrics. I'd be happy to implement that if we think that's better.

### Tests

I extended the `test_aggregate_and_send_metrics` test with variant data instead of adding a new test. Because variant metrics now are a part of metrics, that felt reasonable. However, we can create a separate test if that's preferable.

### Counting variants on the feature level

The variant counting is currently handled on the feature (`Feature` class) level instead of on the variant (`Variant` class) level.
This was done for a few reasons:
- most importantly: we want to count stats for the disabled variant. Because the disabled variant isn't added as part of the `Feature`'s variants, we'd have to count this separately if we placed counts on the `Variant` level.
- `yes_count` and `no_count` are already tracked on the feature, so it felt natural to count variants there too
- it simplifies serialization and the `reset_stats` method

### Casting `DISABLED_VARIANT["name"]`

Described in the [inline comment](https://github.com/Unleash/unleash-client-python/pull/253#discussion_r1205484173):

This cast appears to be the only way to get the linter to accept that we get a string here. As far as I can tell, "name" should always exist and will always be a string. However, the disabled variant isn't typed in any particular way, so it just shows up as a Dict[str, object].

I'd rather avoid this cast if I can, but I don't know if it's possible without re-architecting significant portions of the code and introducing breaking changes. Input would be much appreciated.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Unit tests
- [ ] Spec Tests

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes (the ETAG test is failing, but it fails on main too)
- [ ] Any dependent changes have been merged and published in downstream modules (I don't think there are any downstream modules?)

### DIFF
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -36,6 +36,8 @@ class Feature:
         # Stats tracking
         self.yes_count = 0
         self.no_count = 0
+        ## { [ variant name ]: number }
+        self.variant_counts = {}
 
     def reset_stats(self) -> None:
         """
@@ -57,6 +59,15 @@ class Feature:
             self.yes_count += 1
         else:
             self.no_count += 1
+
+    def count_variant(self, variant_name: str) -> None:
+        """
+        Count a specific variant.
+
+        :param variant_name:
+        :return:
+        """
+        self.variant_counts[variant_name] = self.variant_counts.get(variant_name, 0) + 1
 
     def is_enabled(
         self, context: dict = None, default_value: bool = False
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -47,6 +47,7 @@ class Feature:
         """
         self.yes_count = 0
         self.no_count = 0
+        self.variant_counts = {}
 
     def increment_stats(self, result: bool) -> None:
         """
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -61,7 +61,7 @@ class Feature:
         else:
             self.no_count += 1
 
-    def count_variant(self, variant_name: str) -> None:
+    def _count_variant(self, variant_name: str) -> None:
         """
         Count a specific variant.
 
@@ -117,5 +117,5 @@ class Feature:
             except Exception as variant_exception:
                 LOGGER.warning("Error selecting variant: %s", variant_exception)
 
-        self.count_variant(cast(str, variant["name"]))
+        self._count_variant(cast(str, variant["name"]))
         return variant
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -65,7 +65,7 @@ class Feature:
         """
         Count a specific variant.
 
-        :param variant_name:
+        :param variant_name: The name of the variant to count.
         :return:
         """
         self.variant_counts[variant_name] = self.variant_counts.get(variant_name, 0) + 1
--- a/UnleashClient/features/Feature.py
+++ b/UnleashClient/features/Feature.py
@@ -117,4 +117,5 @@ class Feature:
             except Exception as variant_exception:
                 LOGGER.warning("Error selecting variant: %s", variant_exception)
 
+        self.count_variant(variant["name"])
         return variant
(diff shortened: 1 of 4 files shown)


UnleashClient/features/Feature.py
```diff
@@ -105,4 +117,5 @@ def get_variant(self, context: dict = None) -> dict:
             except Exception as variant_exception:
                 LOGGER.warning("Error selecting variant: %s", variant_exception)
 
+        self.count_variant(cast(str, variant["name"]))
```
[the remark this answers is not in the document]
sighphyre: It's a faux cast, this doesn't actually change the underlying code and it only affects mypy. But I do agree with both your points - I'd both rather avoid it and don't see a way around it without changing the actual type.

UnleashClient/features/Feature.py
```diff
@@ -58,6 +61,15 @@ def increment_stats(self, result: bool) -> None:
         else:
             self.no_count += 1
 
+    def count_variant(self, variant_name: str) -> None:
```
sighphyre, line 64: This looks like an internal method that should only be called by the SDK itself, so it might be better off being a non public method. That being said, we don't do that anywhere else in the file so I won't fuss too much if you disagree

thomasheartman: Yeah, that's totally fair. I agree that that might be better. Does python have a way to make anything non-public? The closest might be using the dunderscore, i.e:

```suggestion
    def __count_variant(self, variant_name: str) -> None:
```

tests/unit_tests/test_features.py
```diff
@@ -76,3 +76,31 @@ def test_select_variation_variation(test_feature_variants):
...
     assert selected_variant["enabled"]
     assert selected_variant["name"] == "VarB"
+
+
+def test_variant_metrics_are_reset(test_feature_variants):
+    test_feature_variants.get_variant({"userId": "2"})
+    assert test_feature_variants.variant_counts["VarB"] == 1
+
+    test_feature_variants.reset_stats()
+    assert test_feature_variants.variant_counts == {}
```
sighphyre, line 86: Empty dicts are truthy so this should work. I'm more used to seeing this kind of thing in Python but if you prefer the comparison to empty dict for clarity that's fine too!

```suggestion
    assert not test_feature_variants.variant_counts
```

sighphyre (approved):
LGTM, small feedback, mostly opinions. Up to you to take or discard

thomasheartman:
For sure! It's much appreciated. Been a while since I touched python, so I'm a little out of touch 😅

UnleashClient/features/Feature.py
sighphyre, line 64: Can I suggest this rather? Python doesn't really have a real way of making things private, the dunderscore does _things_ which make it a lot harder to access but not impossible, just a nuisance. The single underscore is enough to signal to developers that they probably shouldn't touch it unless they know what they're doing
```suggestion
    def _count_variant(self, variant_name: str) -> None:
```

sighphyre: Can't edit the suggestion directly so I made a new one!

thomasheartman: Yeah, that's cool! I'll take care of it! 🙋🏼

thomasheartman: Addressed in 9f1ec633a13cf3ce698a9658897f0b9d5e6cc124

ivanklee86:
lgtm and thanks for the clear explanation!  Let me look into those unit tests failures, something is odd. 🤔

PR #256, opened 2023-06-06
Count metrics for non-existing features

thomasheartman:
## Description

This PR is the follow-up to #253 and makes the client track metrics for features that don't exist.

It achieves this by modifying the Unleash client's `is_enabled` and `get_variant` methods:
Whenever they are called with a feature that doesn't exist, we create a new feature with that name, `enabled = False`, no strategies, and no variants. The feature gets added to the clients list of features, and is then queried for its state (to track metrics; we know it'll be false).

The tests ensure that metrics are sent for these nonexistent features, but does not care about the implementation, so we should be able to change that as much as we want later.

### Discussion point

I've been back and forth a couple of times regarding how best to track these features. Because the `send_metrics` function already takes care of everything related to sending and resetting metrics, it'd be nice to leave that as it is. As such, it'd be nice to track nonexistent features as part of the regular list of features. 

The current implementation uses a boolean field to indicate whether a feature exists only to track metrics or whether it should be evaluated properly. I considered using a subclass to tell them apart instead, but because the loader (`loader.py`) simply **updates** the objects instead of replacing them wholesale (probably to preserve metrics), then that caused some problems.

The current implementation *works*, but I'm not sure it's very pythonic or the best (or even a particularly good) solution, so I'd appreciate any input around that.

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -351,7 +351,9 @@ class UnleashClient:
                 feature_check = feature.is_enabled(context)
 
                 if feature.only_for_metrics:
-                    return self._get_fallback_value(fallback_function, feature_name, context)
+                    return self._get_fallback_value(
+                        fallback_function, feature_name, context
+                    )
 
                 try:
                     if self.unleash_event_callback and feature.impression_data:
@@ -391,7 +393,9 @@ class UnleashClient:
                 # Use the feature's is_enabled method to count the call
                 new_feature.is_enabled(context)
 
-                return self._get_fallback_value(fallback_function, feature_name, context)
+                return self._get_fallback_value(
+                    fallback_function, feature_name, context
+                )
 
         else:
             LOGGER.log(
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
LGTM

UnleashClient/__init__.py
```diff
@@ -449,7 +463,13 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
                     "Error checking feature flag variant: %s",
                     excep,
                 )
-                return DISABLED_VARIATION
+
+                # The feature doesn't exist, so create it to track metrics
+                new_feature = Feature.metrics_only_feature(feature_name)
+                self.features[feature_name] = new_feature
+
+                # Use the feature's is_enabled method to count the call
```
thomasheartman, line 471:
```suggestion
                # Use the feature's get_variant method to count the call
```

PR #257, opened 2023-06-06
chore: remove duplicated test

thomasheartman:
This test appears to be the **exact same** as the one right above
it (`test_uc_server_error`): copying one into the other triggers no
changes from git.

Unless I'm missing something, I suggest we remove it because it's both
confusing to have two identical tests, and because it takes 20 seconds
to sleep.

They also appear to have been the same since they were introduced in 5c9146219fd0 (almost 4 years ago).

## Type of change

- [x] test cleanup


sighphyre (approved):
LGTM

PR #265, opened 2023-07-28
feat: strategy variants

andreas-unleash:
# Description

Implements strategy variants based on Client Specification v4.3.0

Strategy variants take priority over feature variants.  A feature variant (if present) will be  returned if one is not found in the enabled strategy

Fixes # (issue)

## Type of change

- [1] New feature (non-breaking change which adds functionality)

- [1 ] This change requires a documentation update

# How Has This Been Tested?

- [x] Unit tests
- [1] Spec Tests

# Checklist:

- [1 ] My code follows the style guidelines of this project
- [1] I have performed a self-review of my own code
- [ ] My changes generate no new warnings


UnleashClient/constraints/Constraint.py
```diff
@@ -226,10 +226,9 @@ def apply(self, context: dict = None) -> bool:
                     constraint_check = self.check_semver_operators(
                         context_value=context_value
                     )
-            else:
-                # This is a special case in the client spec - so it's getting it's own handler here
-                if self.operator is ConstraintOperators.NOT_IN:  # noqa: PLR5501
-                    constraint_check = True
+            # This is a special case in the client spec - so it's getting it's own handler here
+            elif self.operator is ConstraintOperators.NOT_IN:  # noqa: PLR5501
+                constraint_check = True
```
andreas-unleash, line 231: Linter was failing

thomasheartman:
Regarding the failing tests:

> FAILED tests/unit_tests/api/test_feature.py::test_get_feature_toggle_etag_present - assert '' == 'W/"730-v0ozr...7rQ5PxkXfjYQ"'

This one I expected to see. I ran into the same one when I was doing python a little while back. @sighphyre and I both couldn't figure out why it fails, but it appears to have been a breaking change in one of our dependencies. What we did then was just to say that this test should be ignored until some date. You **could** just extend that for a bit, but we should really look into it.

> FAILED tests/unit_tests/test_client.py::test_uc_registers_variant_metrics_for_nonexistent_features - TypeError: the JSON object must be str, bytes or bytearray, not NoneType

Now this one worries me a bit more. This is what I was working on previously. It didn't have any issues back then, so it's strange that it should fail now. Does this test also fail on main?

andreas-unleash:
> Regarding the failing tests:
> 
> > FAILED tests/unit_tests/api/test_feature.py::test_get_feature_toggle_etag_present - assert '' == 'W/"730-v0ozr...7rQ5PxkXfjYQ"'
> 
> This one I expected to see. I ran into the same one when I was doing python a little while back. @sighphyre and I both couldn't figure out why it fails, but it appears to have been a breaking change in one of our dependencies. What we did then was just to say that this test should be ignored until some date. You **could** just extend that for a bit, but we should really look into it.
> 
> > FAILED tests/unit_tests/test_client.py::test_uc_registers_variant_metrics_for_nonexistent_features - TypeError: the JSON object must be str, bytes or bytearray, not NoneType
> 
> Now this one worries me a bit more. This is what I was working on previously. It didn't have any issues back then, so it's strange that it should fail now. Does this test also fail on main?

Thanks for the info on the first one.  The second is around the area I have been working, so I will look into it

UnleashClient/deprecation_warnings.py
```diff
@@ -10,7 +10,7 @@ def strategy_v2xx_deprecation_check(strategies: list) -> None:
     for strategy in strategies:
         try:
             # Check if the __call__() method is overwritten (should only be true for custom strategies in v1.x or v2.x.
-            if strategy.__call__ != Strategy.__call__:
+            if strategy.__call__ != Strategy.__call__:  # type:ignore
```
thomasheartman, line 13: Is this linter stuff as well?

UnleashClient/strategies/Strategy.py
```diff
@@ -24,9 +33,13 @@ def __init__(
         parameters: dict = {},
         segment_ids: list = None,
         global_segments: dict = None,
+        variants: list = None,
     ) -> None:
+        if variants is None:
+            variants = []
         self.parameters = parameters
         self.constraints = constraints
+        self.variants = variants
```
thomasheartman, line 42: Just to remain stylistically consistent with the other arguments here, how about this?

```suggestion
        self.parameters = parameters
        self.constraints = constraints
        self.variants = variants or []
```

thomasheartman (approved):
The changes here look good to me 😄 As long as the tests pass and the client spec is updated, then that's good! I'm guessing that the cases where you get the old variants are the same as they were before, so we don't need new tests for that?

UnleashClient/features/Feature.py
```diff
@@ -110,19 +105,53 @@ def get_variant(self, context: dict = None) -> dict:
...
                 LOGGER.warning("Error selecting variant: %s", variant_exception)
 
+        if variant is None:
+            variant = copy.deepcopy(DISABLED_VARIATION)
+
         self._count_variant(cast(str, variant["name"]))
         return variant
 
+    def _get_evaluation_result(
+        self, context: dict = None, default_value: bool = False
```
sighphyre, line 134: `default_value` doesn't seem to be used by this function

UnleashClient/features/Feature.py
```diff
@@ -110,19 +105,53 @@ def get_variant(self, context: dict = None) -> dict:
...
+        is_feature_enabled = False
+
+        if evaluation_result is not None:
+            variant = evaluation_result.variant
+            is_feature_enabled = evaluation_result.enabled
+
+        if (
+            is_feature_enabled
+            and self.variants is not None
+            and (variant is None or variant == DISABLED_VARIATION)
```
sighphyre, line 119: Do we get back the disabled variant or a non variant here?  From a look over the code, the None check here shouldn't ever happen

UnleashClient/strategies/Strategy.py
sighphyre: Good suggestion, has the nice side effect of defaulting this to an empty list if someone passes hot garbage like `False` here and isn't using a static type checker

tests/unit_tests/test_features.py
```diff
@@ -1,14 +1,26 @@
...
 from UnleashClient.variants import Variants
 
 (EMAIL_LIST, CONTEXT) = generate_email_list(20)
 
+BASE_FLEXIBLE_ROLLOUT_DICT = {
+    "name": "flexibleRollout",
+    "parameters": {"rollout": 50, "stickiness": "userId", "groupId": "AB12A"},
+    "variants": VARIANTS_WITH_STICKINESS,
+    "constraints": [
+        {"contextName": "environment", "operator": "IN", "values": ["staging", "prod"]},
```
sighphyre, line 17: I think the constraints here are noise, we should be able to do the same test with rollout 0/100 and have it a lot clearer without digging into the weeds of constraints if something goes wrong and needs debugging

UnleashClient/deprecation_warnings.py
sighphyre: This is the type checker, it's gotten a little confused and can't correctly infer the type of `Strategy`. I think this is okay

UnleashClient/strategies/Strategy.py
andreas-unleash: Nice, sure

UnleashClient/features/Feature.py
```diff
@@ -83,19 +85,12 @@ def is_enabled(
...
-                    strategy_result = any(x.execute(context) for x in self.strategies)
-                else:
-                    # If no strategies are present, should default to true. This isn't possible via UI.
-                    strategy_result = True
+        flag_value = False
 
-                flag_value = strategy_result
-            except Exception as strategy_except:
-                LOGGER.warning("Error checking feature flag: %s", strategy_except)
+        if evaluation_result is not None:
```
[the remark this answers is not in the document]
andreas-unleash: good point - cleaning it up

UnleashClient/strategies/Strategy.py
```diff
@@ -66,6 +88,12 @@ def parsed_constraints(self) -> Iterator[Constraint]:
             for constraint in segment["constraints"]:
                 yield Constraint(constraint_dict=constraint)
 
+    @property
+    def parsed_variants(self) -> Variants:
+        return Variants(
+            variants_list=self.variants, group_id=self.parameters.get("group_id", "")
```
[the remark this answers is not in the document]
andreas-unleash: Not sure default stickiness is right here - It is the feature_name when called from Feature.  Would a random string fit better here?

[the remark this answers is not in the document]
sighphyre: Oh good point actually!

Hmm, now that I think about it, can group_id actually even be missing? Since this goes through the new flow, this should always be set, right? Maybe we don't need the default value at all

UnleashClient/features/Feature.py
andreas-unleash: it was marked as deprecated

sighphyre: We should do something with it though, if we don't want to include it we can take it out, if we do then we should actually use the value. Seems weird to have a function that takes a parameter and immediately discards it

andreas-unleash: Agreed, will remove it

UnleashClient/constants.py
```diff
@@ -6,7 +6,7 @@
 REQUEST_TIMEOUT = 30
 REQUEST_RETRIES = 3
 METRIC_LAST_SENT_TIME = "mlst"
-CLIENT_SPEC_VERSION = "4.2.2"
+CLIENT_SPEC_VERSION = "4.3.0"
```
sjaanus, line 9: We could might aswell put it to 4.3.1

tests/specification_tests/test_client_specs.py
```diff
@@ -77,6 +91,6 @@ def test_spec(spec):
         toggle_name = test_data["toggleName"]
         expected = test_data["expectedResult"]
         context = test_data.get("context")
-
+        print("Test-data context: ", context)
```
sjaanus, line 94: Can get rid of this probably?

UnleashClient/features/Feature.py
sighphyre: I see a rereview has been requested but I don't think this one has been done yet

UnleashClient/constants.py
sighphyre, line 9: Good suggestion by @sjaanus, just making this easier to do

```suggestion
CLIENT_SPEC_VERSION = "4.3.1"
```

sighphyre (commented):
Still waiting on a patch to the default property that isn't used. Outside of that I'm happy to approve, the null handling looks substantially saner, nice one!

UnleashClient/features/Feature.py
andreas-unleash: Sorry about that - must have missed it - Removed

sighphyre: All good, story of my life!

sighphyre (approved):
LGTM!

UnleashClient/features/Feature.py
```diff
@@ -110,19 +95,46 @@ def get_variant(self, context: dict = None) -> dict:
...
         self._count_variant(cast(str, variant["name"]))
         return variant
 
+    def _get_evaluation_result(self, context: dict = None) -> EvaluationResult:
+        strategy_result = EvaluationResult(False, None)
+        if self.enabled:
+            try:
+                if self.strategies:
+                    enabled_strategy: Strategy = next(
+                        (x for x in self.strategies if x.execute(context)), None
```
FredrikOseberg, line 122: Just a note: I really hate this shorthand

PR #274, opened 2023-10-05
Dependent features

Tymek:
# Description

Related to https://github.com/Unleash/unleash/issues/2255

Introducing concept of "dependency" on a feature toggle. Feature toggle, in order to be enabled an return a variant, can require the specified state of another feature toggle. There will be a `dependencies` property on each feature.

```
{
 ...
 "dependencies": {
   "feature": "string-featureId",
   "enabled: true, // boolean, state that is expected defaults to `true`
   "variants": ["variant1", "variant2"] // optional, variants allowed for feature to be enabled
 }
}
```

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:


tests/unit_tests/api/test_feature.py
```diff
@@ -90,7 +90,7 @@ def test_get_feature_toggle_failed_etag():
 
 
 @pytest.mark.skipif(
-    date.today() < date(2023, 10, 1),
+    date.today() < date(2023, 11, 1),
```
Tymek, line 93: I create this issue about it https://github.com/Unleash/unleash-client-python/issues/275

kwasniew (commented):
Looks good to me however I'm not a Python dev so would appreciate some Pythonista eyes on it :) I'd also add a test showing that we don't record a metrics for the parent feature.

tests/unit_tests/test_client.py
```diff
@@ -355,6 +356,26 @@ def test_uc_not_initialized_isenabled():
...
+    responses.add(
+        responses.GET,
+        URL + FEATURES_URL,
+        json=MOCK_FEATURE_WITH_DEPENDENCIES_RESPONSE,
+        status=200,
+    )
+    responses.add(responses.POST, URL + METRICS_URL, json={}, status=202)
+
+    unleash_client.initialize_client()
+    time.sleep(1)
```
sighphyre, line 371: If we gotta then we gotta but I'd like to avoid tests that use sleep like this. It's making the whole suite crazy slow.  There's some examples in `tests/specification_tests/test_client_specs.py` where we use bootstrapping over responses to achieve the same thing, which avoids the sleep

sighphyre (changes requested):
Overall LGTM, I've left some thoughts, which are optional but should be considered, so I'm going to mark this as changes requested so this doesn't get accidentally merged

UnleashClient/__init__.py
```diff
@@ -484,6 +490,57 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
...
+        """
+        Checks a single feature dependency.
+        """
+
+        dependency_name = dependency["feature"]
+
+        dependency_feature = self.features[dependency_name]
+
+        if not dependency_feature:
+            LOGGER.warning("Feature dependency not found. %s", dependency_name)
```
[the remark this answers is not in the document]
sighphyre: Not a thing in the Python SDK at the moment, unfortunately. Not against adding it, but that should probably be a different PR

UnleashClient/features/Feature.py
```diff
@@ -75,20 +82,20 @@ def _count_variant(self, variant_name: str) -> None:
         """
         self.variant_counts[variant_name] = self.variant_counts.get(variant_name, 0) + 1
 
-    def is_enabled(self, context: dict = None) -> bool:
+    def is_enabled(self, context: dict = None, skip_stats: bool = False) -> bool:
```
Tymek, line 85: Is there a more elegant way to do it? Without passing `skip_stats` around.

sighphyre: Can't think of a way without rearchitecting this a little. It should be possible to move the count metrics methods to public ones and only call those in the is_enabled in the UnleashClient but... I'm not sure that's worth it. This is all internal at the moment and when Yggdrasil arrives here this will go away anyway.

I'm okay with this as it stands. I think we can do better but I also don't think it's worth it to do so. I think this is a reasonable tradeoff

sighphyre (approved):
Thumbs up

Issue #275, opened 2023-10-06
ETag test failing


Tymek:
**Describe the bug**
ETag test in `tests/unit_tests/api/test_feature.py:92` is broken.

**To Reproduce**
1. Run tests

**Additional context**
It's skipping this test was pushed forward in time multiple times. Let's track this issue and attempts to solve it here.

yjabri:
I believe the problem is the argument `json={}` in the [`responses.add(...)`](https://github.com/Unleash/unleash-client-python/blob/main/tests/unit_tests/api/test_feature.py#L81.). According to [RFC 7232 - 4.1 304 Not Modified](https://datatracker.ietf.org/doc/html/rfc7232#section-4.1)
>A 304 response cannot contain a message-body; it is always terminated by the first empty line after the header field

I think that's why calling `session.get` in the test raises
```
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(2 bytes read, -2 more expected)', IncompleteRead(2 bytes read, -2 more expected))
```

I validated that removing the argument and changing the mock to
```
responses.add(
    responses.GET, PROJECT_URL, status=304, headers={"etag": ETAG_VALUE}
)
```
makes the test pass.

Tymek:
@yjabri If you confirmed that it will make tests pass, PRs are welcome 👍🏻
You can try tests in CI with a PR Draft

yjabri:
@Tymek Does this PR look okay to you? I believe it addresses the problem raised.

PR #277, opened 2023-10-09
fix: update responses mock for etag test

yjabri:
# Description

It was brought up in https://github.com/Unleash/unleash-client-python/issues/275 that a test was failing. The reason appears to be because the response was passing `{}` back in a 304 which requests doesn't like. This behavior is consistent with [RFC 7232 - 4.1 304 Not Modified](https://datatracker.ietf.org/doc/html/rfc7232#section-4.1) i.e. a 304 should not have a body. Another way to confirm this is the root cause is to note the exception

```
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(2 bytes read, -2 more expected)', IncompleteRead(2 bytes read, -2 more expected))
```

points out it read two bytes (`{}`) but expected nothing.

Fixes https://github.com/Unleash/unleash-client-python/issues/275

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:

### DIFF
--- a/tests/unit_tests/api/test_feature.py
+++ b/tests/unit_tests/api/test_feature.py
@@ -1,6 +1,3 @@
-from datetime import date
-
-import pytest
 import responses
 from pytest import mark, param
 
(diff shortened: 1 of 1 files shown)


thomasheartman (approved):
Ah, this is great! 🙌🏼 I'm a little surprised at how simple it really was, but I guess we just didn't dig deep enough earlier 😅  Looks good to me!

Tymek:
Thanks you very much for contributing! Since it's related to testing only new release is not needed. Happy Hacktoberfest! :D

PR #282, opened 2023-10-30
fix: update variants to use a seed value for their hash

sighphyre:
This forces the variant distribution to use a seeded hash, rather than falling back to the default of 0.

This fixes an issue in the variant distribution when gradual rollout is at play, since that narrows down the potential buckets before variants kick in. Previously, we'd see very non random distributions of variants in those cases. This change allows for a more statistically sane distribution


gastonfournier (approved):
LGTM

PR #285, opened 2023-11-06
feat: add is_feature_enabled to variant response

### LINKED ISSUE #284: `UnleashClient().get_variant()` returns the same result when the feature is disabled vs enabled and there are no variants defined
**Is your feature request related to a problem? Please describe.**
It's unclear how to differentiate two cases:
- when the feature is enabled, but there are no variants defined for the feature and for the activated strategy
- when the feature is disabled

In both cases `DISABLED_VARIATION` is returned (`{"name": "disabled", "enabled": False}`).

**Describe the solution you'd like**
Could include "is_feature_enabled" field in the return value of `UnleashClient().get_variant()` and `Feature().get_variant()`

**Describe alternatives you've considered**
Now we either have to be aware if there are variants for a specific feature (and also for an activated strategy) or not or call both
`UnleashClient().get_variant()` and `UnleashClient().is_enabled()`, but that messes up the counting. 

**Additional context**
Seems that in versions before 5.8.0 the result used to be `{"name": "disabled", "enabled": True}` if the feature is _enabled_ but there is no variant.

nunogois:
# Description

Adds a `is_feature_enabled` property to the variant response. This can be especially useful to differentiate between:
 - Feature flag that is **disabled**
 - Feature flag that is **enabled** but has **no variants**

Fixes #284

## Type of change

# How Has This Been Tested?

Added new tests and/or adapted existing tests to assume the new behavior.

- [x] Unit tests
- [x] Spec Tests

# Checklist:


tests/unit_tests/test_features.py
```diff
@@ -151,6 +163,23 @@ def test_strategy_variant_is_returned(test_feature_strategy_variants):
...
+        "appName": "test",
+        "environment": "prod",
+        "customField": "1",
+    }
+    variant = test_feature_no_variants.get_variant(context)
+
+    assert variant == {
+        "enabled": False,
+        "name": "disabled",
+        "is_feature_enabled": True,
```
ivarconr, line 182: Why did we choose `is_feature_enabled` over `feature_enabled`?

nunogois: It's the name suggested in the [issue](https://github.com/Unleash/unleash-client-python/issues/284#issue-1973288429), and I noticed we also have a variable with this name in the [existing codebase](https://github.com/Unleash/unleash-client-python/pull/285/files#diff-6ca66179fe5f5e0a1ecaca00106886de49ea4860733d5009b83a1480c44aa43aR122).
If you prefer `feature_enabled` let me know, I can easily make the change 🙂

UnleashClient/features/Feature.py
```diff
@@ -118,6 +118,8 @@ def get_variant(self, context: dict = None, skip_stats: bool = False) -> dict:
                 LOGGER.warning("Error selecting variant: %s", variant_exception)
         if not skip_stats:
             self._count_variant(cast(str, variant["name"]))
+
+        variant["is_feature_enabled"] = is_feature_enabled
```
sighphyre, line 122: I don't think we need the mutation here, this is just a dict so a comprehension should work here and then we can remove the these two lines
```suggestion
        return {**variant, "is_feature_enabled": is_feature_enabled}
```

UnleashClient/features/Feature.py
```diff
@@ -118,6 +118,8 @@ def get_variant(self, context: dict = None, skip_stats: bool = False) -> dict:
...
         return variant
```
sighphyre, line 123: If we take the above, this is just to make it easier to accept the suggestion without another PR
```suggestion
```

UnleashClient/__init__.py
```diff
@@ -427,13 +428,14 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
         """
         context = context or {}
         context.update(self.unleash_static_context)
+        disabled_variation = copy.deepcopy(DISABLED_VARIATION)
```
sighphyre, line 431: Do we need the copy here? This is a constant so we should never be mutating this

tests/unit_tests/test_client.py
```diff
@@ -399,11 +399,36 @@ def test_uc_get_variant():
...
     variant = unleash_client.get_variant("testVariations", context={"userId": "3"})
     assert variant["name"] == "disabled"
     assert not variant["enabled"]
+    assert not variant["is_feature_enabled"]
+
+    unleash_client.destroy()
+
+
+@responses.activate
+def test_uc_get_variant_feature_enabled_no_variants():
```
sighphyre, line 414: Can we get away with bootstrapping rather than a mocked response? These sleeps are making the test suite insanely slow and I'm trying to be defensive about making that problem worse. There's some examples in the `test_client_specs.py` file where we use bootstrap over a mocked response that removes the need for sleeps and HTTP. If we can't then we can't but I gotta ask!

tests/unit_tests/test_features.py
sighphyre: Don't know that I would have picked this up but now that we're talking, I'm leaning towards "feature_enabled" rather

sighphyre (changes requested):
This looks good to me, I'd like it we can at least address the copy/dict comprehension story before we merge though

The bootstrapping is optional but nice, if it's too much work I'm okay to let that one go

UnleashClient/features/Feature.py
nunogois: Thanks, TIL.

UnleashClient/__init__.py
nunogois: We don't. Addressed in https://github.com/Unleash/unleash-client-python/pull/285/commits/cc102a1c36a6bab2fabfda99e103468b45d1e338

Fun story: This was me being a Python noob and not realizing the `del variant["feature_enabled"]` on my `test_client_specs` was actually affecting the `DISABLED_VARIANT` constant that was being returned, making it so it would no longer have the `feature_enabled` key, which in turn caused future `del variant["feature_enabled"]` to fail, since the key would no longer be present. I noticed we `deepcopied` the DISABLED_VARIANT in another place, so I tried that and it worked. 

Overall I think it's better to use the dict comprehension like you've shown me on the other comment, so I adapted the test. I also got rid of the other copies as I don't think they are necessary.

tests/unit_tests/test_client.py
nunogois: Addressed in https://github.com/Unleash/unleash-client-python/pull/285/commits/7c687f0407dca73d47af776d805f0b682fb0bfff

Might be worth doing the same for the rest of the tests, but out of scope for this PR.

UnleashClient/__init__.py
sighphyre: Oooo! Nice spot! Burned by state mutations xD

tests/unit_tests/test_client.py
sighphyre: I'd love to but gosh it's a lot of work. Definitely something on my back burner for when insomnia strikes again

sighphyre (approved):
Love it!

PR #302, opened 2024-03-05
fix:mimesis pinned to 11.1.0

gardleopard:
WIP


thomasheartman (approved):
Looks like a sensible pin to me. It's only for testing anyway, so seems like a low-risk change.

PR #304, opened 2024-03-07
fix: correctly count metrics when a flag with dependencies has disabled parents

thomasheartman:
# Description

This PR fixes a bug where we didn't count metrics correctly: if the dependencies were not satisfied during a `get_variant` call, we wouldn't track any metrics for the flag at all. During an `is_enabled` call, we'd track the result of the child flag's evaluation without taking into account the dependencies.

Note that the **behavior** was still correct. This is only about metrics.

The root cause is related to how we check dependencies and act on that result: the Unleash client checks dependencies as part of its `is_enabled` and `get_variant` calls, but the metrics are tracked in the individual `Feature` objects.

The current implementation was the quickest way I could fix the issue. I think we should restructure it (probably; either now or later) to avoid calling internal methods of the Feature object. However, it serves as a nice way to show what's wrong and how we can fix it. (Maybe just make that `_count_variant` method non-internal?)

Due to the way the code is organized already, it might make sense to move the dependencies check into the Feature class instead of performing it in the Unleash Client. However, that would require us to have some way to access those dependencies from the dependent feature. This probably means more restructuring. I'm not sure that's worth it. Maybe it's better to leave as is? I'll take any input on it 🙋🏼 

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:

- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -364,7 +364,9 @@ class UnleashClient:
         if self.unleash_bootstrapped or self.is_initialized:
             try:
                 feature = self.features[feature_name]
-                dependency_check = self._dependencies_are_satisfied(feature_name, context)
+                dependency_check = self._dependencies_are_satisfied(
+                    feature_name, context
+                )
 
                 if dependency_check:
                     feature_check = feature.is_enabled(context)
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -364,9 +364,13 @@ class UnleashClient:
         if self.unleash_bootstrapped or self.is_initialized:
             try:
                 feature = self.features[feature_name]
-                feature_check = feature.is_enabled(
-                    context
-                ) and self._dependencies_are_satisfied(feature_name, context)
+                dependency_check = self._dependencies_are_satisfied(feature_name, context)
+
+                if dependency_check:
+                    feature_check = feature.is_enabled(context)
+                else:
+                    feature.increment_stats(False)
+                    feature_check = False
 
                 if feature.only_for_metrics:
                     return self._get_fallback_value(
@@ -449,6 +453,8 @@ class UnleashClient:
                 feature = self.features[feature_name]
 
                 if not self._dependencies_are_satisfied(feature_name, context):
+                    feature.increment_stats(False)
+                    feature._count_variant("disabled")
                     return DISABLED_VARIATION
 
                 variant_check = feature.get_variant(context)
--- a/tests/unit_tests/test_client.py
+++ b/tests/unit_tests/test_client.py
@@ -572,6 +572,7 @@ def test_uc_doesnt_count_metrics_for_dependency_parents(unleash_client):
     time.sleep(1)
 
     child = "ChildWithVariant"
+    parent = "Parent"
     # Check a flag that depends on a parent
     unleash_client.is_enabled(child)
     unleash_client.get_variant(child)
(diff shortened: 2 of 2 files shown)


tests/unit_tests/test_client.py
```diff
@@ -572,6 +572,7 @@ def test_uc_doesnt_count_metrics_for_dependency_parents(unleash_client):
     time.sleep(1)
 
     child = "ChildWithVariant"
+    parent = "Parent"
```
thomasheartman, line 575: Sorry: this had just been bugging me since last night.

UnleashClient/__init__.py
```diff
@@ -449,6 +455,8 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
                 feature = self.features[feature_name]
 
                 if not self._dependencies_are_satisfied(feature_name, context):
+                    feature.increment_stats(False)
+                    feature._count_variant("disabled")
```
thomasheartman, line 459: We never counted stats for variants at all if dependencies were not satisfied.

UnleashClient/__init__.py
```diff
@@ -364,9 +364,15 @@ def is_enabled(
...
-                ) and self._dependencies_are_satisfied(feature_name, context)
+                dependency_check = self._dependencies_are_satisfied(
+                    feature_name, context
+                )
+
+                if dependency_check:
+                    feature_check = feature.is_enabled(context)
+                else:
+                    feature.increment_stats(False)
+                    feature_check = False
```
thomasheartman, line 375: If the flag is enabled itself, but its dependencies are not satisfied, we'd count it as a "yes" previously (by checking "is_enabled" first.

tests/unit_tests/test_client.py
```diff
@@ -581,7 +582,37 @@ def test_uc_doesnt_count_metrics_for_dependency_parents(unleash_client):
...
+    time.sleep(1)
+
+    child = "WithDisabledDependency"
+    parent = "Disabled"
+    # Check a flag that depends on a disabled parent
+    unleash_client.is_enabled(child)
+    unleash_client.get_variant(child)
+
+    # Verify that the parent doesn't have any metrics registered
+    time.sleep(12)
```
sighphyre, line 611: I think this is my only objection here. These tests are very painful to run since they all trigger long sleeps. This one adds a full 13 seconds

The client is being newed up with a 10 second metrics send interval. Can we see what happens to the test suite if we reduce that to 1 and trim the sleep down to say 2 seconds?

thomasheartman: Yeah, sure. I'll have a look. Think I might have had a look before, but can't remember why I didn't follow through in that case. Will check.

thomasheartman: And of course, we can move it out into a separate PR too. But just wanted some input on it first.

sighphyre: > Yeah, sure. I'll have a look. Think I might have had a look before, but can't remember why I didn't follow through in that case. Will check.

I think I have too and failed to improve this a few times. Looks like the build server hates that change. I think these are accidentally doing more than just pausing execution. 

Okay, let's not do that then, this looks to be bigger than just this PR, I just to make sure it wasn't an easy fix we weren't doing.

Let's remove the commit, I'm okay for this to go in without the timing change

thomasheartman: Summary of in-person chat: I think I have a way to fix this. Will take those changes into a new PR (#305 tentatively)

PR #306, opened 2024-03-07
chore: update change log for v5.11.2

thomasheartman:
Add a note about the bug fix that was done in the past two days.


sighphyre (approved):
LGTM

PR #314, opened 2024-06-28
chore: exclude tests from build output

sighphyre:
# Description

This forces the wheel and tarball builds to exclude the test folder in the output. For two reasons:

1) This conflicts with end users test files sometimes
2) I want a reasonable way to exclude sample applications (incoming) from the build output

## Type of change

None of the standard options, just a build fix

# How Has This Been Tested?

Folder diffing, checked the build outputs and the only change we see is that the tests are excluded and the manifest is patched to reflect this

# Checklist:

- [ x ] I have performed a self-review of my own code
- [ x ] My changes generate no new warnings
- [ x ] New and existing unit tests pass locally with my changes


ivarconr (approved):

gardleopard (approved):
LGTM

PR #315, opened 2024-06-28
chore: adapt ruff call in build to new format

sighphyre:
# Description

Ruff got an upgrade, the old `ruff <path>` command we use in the build no longer works. Now apparently it wants `ruff check <path>`

This new version of Ruff also raises a new lint, which is fixed in this PR.

Bonus points for fixing Ruff config in pyproject.toml that had been raising some warnings for some time

Fixes # (issue)

The build

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)\

# How Has This Been Tested?

This PR + build pipeline

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ fmt:
 
 lint:
 	black . --check && \
-	ruff UnleashClient tests docs && \
+	ruff check UnleashClient tests docs && \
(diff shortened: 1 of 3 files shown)


gardleopard (approved):
LGTM

PR #322, opened 2024-11-11
chore: swap core for yggdrasil

sighphyre:
Swaps the core of the SDK for the [Python bindings to Yggdrasil](https://github.com/Unleash/yggdrasil/tree/main/python-engine). 

Currently this "works on my machine" because the engine isn't yet published, but all tests do pass with a local link.

## Details

- Feature and variant evaluations are all handled through Yggdrasil now 
- Metrics counting and retrieval is handled through Yggdrasil now
- The feature loader has been pared down to a much simpler set of functionally, all state diffing is handled by the Rust layer of the engine (currently no diffing but its coming)
- Custom strategies are handled through the Python layer of Yggdrasil now. **_This means an interface change for custom strategies_**
- All hydration paths (bootstrap, load from cache, API hit) now return strings; parsing is handled by the Rust layer of Yggdrasil
- Access to the `features` field has been left intact on the `UnleashClient` class but will now always return empty. This is intended to be swapped out for `feature_definitions`. `features` will be removed in a further clean up PR prior to release

## Discussion points

- The Yggdrasil engine is a bit more fussy about types, specifically on Variants. Yggdrasil returns a dataclass, rather than the untyped dict of the SDK. I've chosen to map this to the legacy format to allow consumers an easier time moving, while giving us an option to move to something a little more concrete at a later point
- Custom strategies is a big change. This is going to require a major version and the new methods in Yggdrasil will error at initialisation time rather than warn at runtime. I think this is okay but I'm open to discussion here
- The engine isn't hinted to be private (and it's not name mangled because that's horrible) because none of the other SDK fields are. I don't believe people should be rooting around with the engine directly but Python also tends to treat its users like responsible adults so I think that's maybe okay
- I've chosen not to handle some of the YddrasilExceptions that can bubble from the engine, specifically for cases where this would be undefined behavior in the Rust layer. These Should Never Happen™ but if they do, it's not really safe for any healthy program to continue what its doing


UnleashClient/__init__.py
```diff
@@ -450,69 +432,51 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
...
-                feature_name,
-            )
-            return DISABLED_VARIATION
+    def _resolve_variant(self, feature_name: str, context: dict) -> dict:
+        """
+        Resolves a feature variant.
+        """
+        variant = self.engine.get_variant(feature_name, context)
+        if variant:
+            return {k: v for k, v in asdict(variant).items() if v is not None}
```
nunogois, line 478: Just curious: Why do we need this?

UnleashClient/__init__.py
```diff
@@ -450,69 +432,51 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
...
+
+        if not variant:
+            if self.unleash_bootstrapped or self.is_initialized:
                 LOGGER.log(
                     self.unleash_verbose_log_level,
-                    "Returning default flag/variation for feature: %s",
+                    "Attempted to get feature flag/variation %s, but client wasn't initialized!",
                     feature_name,
                 )
+            variant = DISABLED_VARIATION
```
nunogois, line 444: Would probably be cleaner to move this part over to `_resolve_variant` and make it always return a variant instead of either a variant or `None`?

UnleashClient/loader.py
```diff
@@ -1,107 +1,23 @@
...
-    global_segments: Optional[dict] = None,
+    engine: UnleashEngine,
 ) -> None:
     """
     Caching
 
     :param cache: Should be the cache class variable from UnleashClient
-    :param feature_toggles: Should be the features class variable from UnleashClient
-    :param strategy_mapping:
+    :param feature_toggles: Should a JSON string containing the feature toggles, equivalent to the response from Unleash API
```
nunogois, line 20:
```suggestion
    :param feature_toggles: Should be a JSON string containing the feature toggles, equivalent to the response from Unleash API
```

UnleashClient/__init__.py
```diff
@@ -312,6 +314,28 @@ def initialize_client(self, fetch_toggles: bool = True) -> None:
                 "Attempted to initialize an Unleash Client instance that has already been initialized."
             )
 
+    def feature_definitions(self) -> dict:
+        """
+        Returns a dict containing all feature definitions known to the SDK at the time of calling.
+        Normally this would be a pared down version of the response from the Unleash API but this
+        can also be as a result of bootstrapping or loading from backup.
```
nunogois, line 321:
```suggestion
        may also result from bootstrapping or loading from a backup.
```

UnleashClient/api/features.py
```diff
@@ -82,7 +82,7 @@ def get_feature_toggles(
         if resp.status_code == 304:
             return None, etag
 
-        return resp.json(), etag
+        return resp.text, etag
```
nunogois, line 85: Shouldn't the method signature response change to `Tuple[str, str]`?

UnleashClient/periodic_tasks/fetch_and_load.py
```diff
@@ -18,9 +20,10 @@ def fetch_and_load_features(
     strategy_mapping: dict,
     request_timeout: int,
     request_retries: int,
+    engine: UnleashEngine,
     project: Optional[str] = None,
 ) -> None:
-    (feature_provisioning, etag) = get_feature_toggles(
+    (feature_toggle_json, etag) = get_feature_toggles(
```
nunogois, line 26: Bit unsure about the name, especially now that it's a str and not a dict. We're not lying, it's still a JSON, but in a string. In other places we used `state`, maybe that could work better here?

nunogois (approved):
Other than a few nitpicks it seems to me like you're doing all the right things 👍

UnleashClient/api/register.py
```diff
@@ -52,7 +52,7 @@ def register_client(
         "interval": metrics_interval,
         "platformName": python_implementation(),
         "platformVersion": python_version(),
-        "yggdrasilVersion": None,
+        "yggdrasilVersion": "0.14.0",
```
ivanklee86, line 55: Can you do something like `yggdrasil.__version__` ?  I think a constant/tying it back to the requirements might be better!

ivanklee86 (approved):
Some random thoughts:
- Should update the development docs on what yggdrasil is/how to upgrade.
- Custom strategy changes probably need updates.

UnleashClient/api/register.py
sighphyre: Yes! I realised this last night actually and I was wondering if someone was going to spot it :laughing: Good spot, will swap it out

UnleashClient/periodic_tasks/fetch_and_load.py
sighphyre: Cheers, good suggestion!

UnleashClient/api/features.py
sighphyre: It does, changed now! On that notes, apparently we don't have MyPy setup either

UnleashClient/__init__.py
sighphyre: It's a good question. Yggdrasil core returns a dataclass here (which is like an object without methods, similar to a Rust struct). I'm a fan of dataclasses because they're quite explicit about their intent. The SDK at the moment returns an untyped dict. So I don't want to change it for 3 reasons:

- Yggdrasil abstractions should stay in Yggdrasil, bleeding internal classes outside of the domain boundary is poor encapsulation
- This is one less thing for consumers to have to think about or worry about, that one is quite important to me in Python because its quite hard to know all the places you may be using a return type. This means no one has to hunt down all their variants and change the way they access data on them
- Dataclasses do optionals, which become nulls. The client spec suggests that variants without payloads shouldn't include that field rather than including a null. The dict comprehension here is filtering out nulls

UnleashClient/__init__.py
sighphyre: I left it alone because I was going to implement the fallback function, which needs to know if there's a variant or not... then I realised we never got around to implementing that. So I'm leaving this until we do actually implement it

sighphyre:
> Some random thoughts:
> 
>     * Should update the development docs on what yggdrasil is/how to upgrade.
> 
>     * Custom strategy changes probably need updates.

Cheers, both are absolutely on my list of things to do. I just wanted to keep this PR as narrow as possible. Will definitely be a thing prior to an actual release

UnleashClient/api/register.py
sighphyre: Fixed in [02bb05](https://github.com/Unleash/unleash-client-python/pull/322/commits/02bb05ec4817c4d68ddcafe7102c1ddbadf97b6f)

PR #324, opened 2024-11-11
Safely handle datetime constraint values with and without timezone info

### LINKED ISSUE #323: Datetime constraints result in TypeError if currentTime is not provided or currentTime is not timezone-aware
**Describe the bug**
A datetime constraint can be created in Unleash with a timezone-aware or timezone-naive ISO datetime string. If a timezone-aware datetime is used (which appears to be the default for the Unleash web app), the Python client will fail when evaluation that constraint if a timezone-naive datetime object/string is provided as the `currentTime` in the context. The same is true of the inverse (tz-naive constraint and tz-aware `currentTime`).  

This is mainly an issue if `currentTime` is not set in the context and the Unleash client provide its own `currentTime` value. If a user providers a `currentTime` value themselves, it's reasonable enough to require them to be consistent (all tz-aware or all tz-naive). But currently, if the default `currentTime` is used, it fails if the constraint is tz-aware. 

**Logs**

```
TypeError: can't compare offset-naive and offset-aware datetimes
```

jacob-indigo:
# Description

The Unleash web app provided timezone info when creating date constraints so this PR updates the client to also provide timezone info for datetime objects it creates. And then gracefully fall back to UTC if an ISO string is received with no timezone info. 

Fixes #323

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/UnleashClient/constraints/Constraint.py
+++ b/UnleashClient/constraints/Constraint.py
@@ -1,5 +1,5 @@
 # pylint: disable=invalid-name, too-few-public-methods, use-a-generator
-from datetime import datetime
+from datetime import datetime, timezone
 from enum import Enum
 from typing import Any, Optional, Union
 
@@ -125,6 +125,11 @@ class Constraint:
         return return_value
 
     def check_date_operators(self, context_value: Union[datetime, str]) -> bool:
+        if isinstance(context_value, datetime) and context_value.tzinfo is None:
+            raise ValueError(
+                "If context_value is a datetime object, it must be timezone (offset) aware."
+            )
+
         return_value = False
         parsing_exception = False
 
@@ -139,8 +144,16 @@ class Constraint:
 
         try:
             parsed_date = parse(self.value)
+
+            # If parsed date is timezone-naive, assume it is UTC
+            if parsed_date.tzinfo is None:
+                parsed_date = parsed_date.replace(tzinfo=timezone.utc)
+
             if isinstance(context_value, str):
                 context_date = parse(context_value)
+                # If parsed date is timezone-naive, assume it is UTC
+                if context_date.tzinfo is None:
+                    context_date = context_date.replace(tzinfo=timezone.utc)
             else:
                 context_date = context_value
         except DateUtilParserError:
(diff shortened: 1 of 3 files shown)


UnleashClient/constraints/Constraint.py
```diff
@@ -139,8 +144,16 @@ def check_date_operators(self, context_value: Union[datetime, str]) -> bool:
 
         try:
             parsed_date = parse(self.value)
+
+            # If parsed date is timezone-naive, assume it is UTC
+            if parsed_date.tzinfo is None:
```
sighphyre, line 149: This originates on the server and I don't think this should ever happen. Is this necessary?

UnleashClient/constraints/Constraint.py
```diff
@@ -197,7 +210,8 @@ def apply(self, context: dict = None) -> bool:
 
             # Set currentTime if not specified
             if self.context_name == "currentTime" and not context_value:
-                context_value = datetime.now()
+                # Use the current system time in the local timezone (tz-aware)
+                context_value = datetime.now(timezone.utc).astimezone()
```
sighphyre, line 214: I think this makes sense!

UnleashClient/constraints/Constraint.py
```diff
@@ -139,8 +144,16 @@ def check_date_operators(self, context_value: Union[datetime, str]) -> bool:
...
             parsed_date = parse(self.value)
+
+            # If parsed date is timezone-naive, assume it is UTC
+            if parsed_date.tzinfo is None:
+                parsed_date = parsed_date.replace(tzinfo=timezone.utc)
+
             if isinstance(context_value, str):
                 context_date = parse(context_value)
+                # If parsed date is timezone-naive, assume it is UTC
+                if context_date.tzinfo is None:
```
sighphyre, line 155: I'm not sure this is the right thing to do here. I'd like both a a `Date` and a `str` to go through the same code path and right now one is raising an error and the other silently changing the underlying data

sighphyre (approved):
Okay I thought about it, this might not be perfect for maintenance reasons but I believe the flow is more or less correct and since this code is going away in the next major, I see no harm in merge + releasing this as a patch

PR #325, opened 2024-11-14
chore: cleanup unused domain

sighphyre:
Removes the now hanging domain objects that are now handled by Yggdrasil

Notably, this removes `features` on the client (replaced by `list_features` in Yggdrasil) and drops the SDK from listing the built in strategies when sending metrics


nunogois (approved):
LGTM

ivanklee86 (approved):

PR #329, opened 2024-11-15
chore: safe context

sighphyre:
This converts all values in the context object into their representative string type before sending it to Yggdrasil.

Also adds the current time property if not set. This was previously set in the constraints code, which no longer exists


gastonfournier (approved):
LGTM!

PR #337, opened 2025-01-07
chore: bump client spec to 5.1.9

sighphyre:
Bumps client spec to 5.1.9 so we can confirm unicode characters work with Yggdrasil on Windows


sighphyre:
This is a problem but probably not a big problem. In theory there should be no way to get non ASCII chars to the SDK since the server doesn't support it, although Gitlab may be doing something else

PR #339, opened 2025-01-08
Chore/bump client spec

sighphyre:
Update client spec to double check that the SDK and/or Yggdrasil handles unicode correctly


gastonfournier (approved):
LG!

PR #345, opened 2025-02-13
chore: remve gh-action slack notify

ivarconr:
# Description

removes the slack notifier for github as we do not use it.


chriswk (approved):
LGTM

PR #342, opened 2025-01-27
feat: client identification headers

Tymek:
# Description

Identification headers consistent with other Unleash SDKs.

> This PR adds standardized client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
> - `x-unleash-appname`: the name of the application that is using the client. `UNLEASH-APPNAME` will be deleted in another PR (expand/contract pattern)
> - `x-unleash-connection-id`: an internal unique identifier for the current instance of the client generated by the built-in crypto lib
> - `x-unleash-sdk`: sdk information in the format `unleash-client-<language>:<version>`
> 
> All the headers are intended for the Unleash team. Changes should be implemented in a way that does not change SDK behavior in a significant way.
> The main use cases we have are:
> * capacity planning by knowing the number of unique connections made to the backend API
> * debugging misconfigured SDKs sending more traffic than expected

Fixes [issue/1-3274](https://linear.app/unleash/issue/1-3274/uniqueness-headers-in-python-sdk)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] Unit tests
- [ ] Spec Tests
- [ ] Integration tests / Manual Tests

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #346, opened 2025-02-27
feat: new headers and payload ( connection id and interval)

sjaanus:
## Request updates

- Fetch flag calls include the `unleash-interval` header.
- Registration calls now have `connectionId` added to the payload (previously, it only had the `unleash-connection-id` header).

## Metrics updates

- Send metrics calls include the `unleash-interval` header.
- Send metrics now have `connectionId` added to the payload (previously, it only had the `unleash-connection-id` header).

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -110,7 +110,7 @@ class UnleashClient:
         self.unleash_app_name = app_name
         self.unleash_environment = environment
         self.unleash_instance_id = instance_id
-        self.connection_id = str(uuid.uuid4())
+        self._connection_id = str(uuid.uuid4())
         self.unleash_refresh_interval = refresh_interval
         self.unleash_request_timeout = request_timeout
         self.unleash_request_retries = request_retries
@@ -187,6 +187,10 @@ class UnleashClient:
     @property
     def unleash_refresh_interval_str_millis(self) -> str:
         return str(self.unleash_refresh_interval * 1000)
+
+    @property
+    def connection_id(self):
+        return self._connection_id
 
     def initialize_client(self, fetch_toggles: bool = True) -> None:
         """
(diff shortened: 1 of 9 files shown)


tests/unit_tests/test_client.py
```diff
@@ -1078,3 +1079,37 @@ def test_identification_headers_unique_connection_id():
         "UNLEASH-CONNECTION-ID"
     ]
     assert connection_id_first_client != connection_id_second_client
+
+
+@responses.activate
+def test_identification_values_are_passed_in():
```
sjaanus, line 1085: This is testing all 3 requests, headers and body

UnleashClient/__init__.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         self.unleash_app_name = app_name
         self.unleash_environment = environment
         self.unleash_instance_id = instance_id
+        self.connection_id = str(uuid.uuid4())
```
kwasniew, line 113: we had connection id before. is this code moved?

tests/unit_tests/test_client.py
kwasniew: I remember we had this test added: https://github.com/Unleash/unleash-client-python/pull/342/files#diff-6fd6de24505791fbd1ad578927a8fdd87d15a14b7d194600d68434ddc3e53e00R1038 Are those 2 tests testing similar things?

kwasniew (approved):

UnleashClient/__init__.py
sjaanus: Yes, it was initialized only in headers before, now we can use it also in body if needed.

https://github.com/Unleash/unleash-client-python/pull/346/files#diff-d8684313c6cfe7bbd96df3385f5b28c7f337cd5159fc08f79b518df75cc13475L218

PR #349, opened 2025-04-25
fix: publish pipeline

sighphyre:


gastonfournier (approved):
LG! Let's try this out

PR #353, opened 2025-07-01
fix: allow anything that implements str as a context property

sighphyre:
Related to https://github.com/Unleash/yggdrasil/issues/257

Allows anything that implements `__str__` to be used as a context property. For complex types like UUIDs, this means geting a reasonable string representation of the value


gastonfournier (approved):
LGTM!

PR #357, opened 2025-07-02
feat: events

### LINKED ISSUE #356: Feature request: implement a subscriber API to align with other SDK offerings
**Is your feature request related to a problem? Please describe.**

We are trying to log toggle status changes so that we know when exactly those changes have been applied to the application, but we have no way to subscribe to those events.

**Describe the solution you'd like**
A solution [similar to the JDK's SDK](https://docs.getunleash.io/reference/sdks/java#subscriber-api) which allows us to create subscribers to different events so that our applications can act upon those events.

**Describe alternatives you've considered**
For now we will probably manually keep track of the status of the toggles and wrap the `is_enabled` call so that when we notice a status change we can print a log. Either that or try to override the `fetch_and_load_features` function somehow.

**Additional context**
N/A

sighphyre:
Adds in events for when features are fetched and when the SDK becomes ready.

This brings us more inline with the other SDKs which have these events at minimum.

Closes #356

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -24,7 +24,7 @@ from UnleashClient.constants import (
     SDK_NAME,
     SDK_VERSION,
 )
-from UnleashClient.events import UnleashEvent, UnleashEventType
+from UnleashClient.events import UnleashEvent, UnleashEventType, UnleashReadyEvent
 from UnleashClient.loader import load_features
 from UnleashClient.periodic_tasks import (
     aggregate_and_send_metrics,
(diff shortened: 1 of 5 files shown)


UnleashClient/events.py
```diff
@@ -11,17 +12,51 @@ class UnleashEventType(Enum):
...
+    """
+    Event indicating that the Unleash client has fetched feature flags.
+    """
+
+    raw_features: str
+
+    @property
+    def features(self) -> dict:
+        if not hasattr(self, "_parsed_payload"):
+            self._parsed_payload = loads(self.raw_features)["features"]
```
sighphyre, line 61: Annoyingly this now becomes part of the public API.

Kinda struggled a little with choosing between only returning the features vs returning the whole response. Leaning ever so slightly in the direction of only returning the features since its less confusing for end users.

That being said, every time we've done that in the past (fetch features, bootstrap, load from backup), it's been a mistake that needs to be patched later. Open to being told this is similar but I can't see how right now

UnleashClient/__init__.py
```diff
@@ -46,6 +51,37 @@
...
 
+def build_ready_callback(
+    event_callback: Optional[Callable[[BaseEvent], None]] = None,
+) -> Optional[Callable]:
+    """
+    Builds a callback function that can be used to notify when the Unleash client is ready.
+    """
+
+    if not event_callback:
+        return None
```
sighphyre, line 62: Poor mans optional.map()

gastonfournier: I wonder if it makes sense to support this optional parameter. IMO, if you're going to call this function, you better have a callback. If you supply something that's not a callback then you should expect an error. But having this ability to call all functions with optional and checking if it's set seems a bit off.

On the other hand, if this function can be called from multiple places it centralizes the check in one single place in the code, so just picking your brain about it.

UnleashClient/periodic_tasks/fetch_and_load.py
```diff
@@ -35,6 +39,15 @@ def fetch_and_load_features(
...
         cache.set(FEATURES_URL, state)
+        if event_callback:
+            event = UnleashFetchedEvent(
+                event_type=UnleashEventType.FETCHED,
+                event_id=uuid.uuid4(),
+                raw_features=state,
+            )
+            event_callback(event)
+        if ready_callback:
+            ready_callback()
```
gastonfournier, line 50: I think I kind of like the idea of removing the if statement and having functions doing nothing.

gastonfournier (approved):
This looks nice, just left you a couple of stylish comments

PR #358, opened 2025-07-04
chore: improve test speed

sighphyre:
This improves the speed of the test suite by leveraging the new events in the SDK. Because we now have a deterministic way of knowing when the SDK is ready to process toggles and we know when the SDK has done fetching, we can both use thread events for more fine grained control and drop the polling interval. Drops the time taken for the test suite from around 80 seconds to about 10 seconds.

## Details

 Almost every test gets the same treatment - instead of setting up a basic UnleashClient, instead it gets a client along with a pair of thread signals that indicate when a fetch or ready event happens. This means we're not blocking the main thread with sleep calls and instead of blindly waiting, we can respond when we know that the SDK is ready/has fetched.

### DIFF
--- a/tests/unit_tests/test_client.py
+++ b/tests/unit_tests/test_client.py
@@ -1152,7 +1152,6 @@ def test_ready_signal_works_with_bootstrapping():
         if kw["data"].event_type == UnleashEventType.READY:
             nonlocal trapped_events
             trapped_events += 1
-            ready_signal = threading.Event()
 
     def example_callback(event: UnleashEvent):
         send_data.send("anonymous", data=event)


tests/unit_tests/test_client.py
```diff
@@ -77,46 +96,105 @@ def cache(tmpdir):
 
 
 @pytest.fixture()
-def unleash_client(cache):
+def readyable_unleash_client(cache):
```
sighphyre, line 99: Pytest fixures have some weird scope closing rules that makes just using the function above not possible. So each of these fixtures gets their own little copy of the code

tests/unit_tests/test_client.py
```diff
@@ -225,12 +304,15 @@ def test_uc_lifecycle(unleash_client):
         status=200,
         headers={"etag": "W/somethingelse"},
     )
-    time.sleep(REFRESH_INTERVAL * 2)
+    fetch_signal.clear()
+    fetch_signal.wait(timeout=REFRESH_INTERVAL * 3)
```
sighphyre, line 308: Gotta wait for the first fetch (304), then the second fetch (200), then we give the lock another increment as a grace period - so REFRESH * 3

sighphyre: I want to say it's clearer when you have the whole test in front of you. I'd like to actually just throw this test away at some point. It's really not a very clear test

tests/unit_tests/test_client.py
```diff
@@ -1273,7 +1399,7 @@ def test_identification_values_are_passed_in():
     except ValueError:
         assert False, "Invalid UUID format in UNLEASH-CONNECTION-ID"
 
-    time.sleep(3)
+    time.sleep(1.5)
```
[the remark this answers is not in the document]
sighphyre: Yessir! This one is waiting for metrics unfortunately

PR #367, opened 2025-08-26
Refactor: connectors abstraction

Tymek:
# Description

Refactor how SDK is getting flag definitions ('fetching modes') in preparation for streaming support.

Helper functions are rolled into "connectors", with 2 implementations available: `OfflineConnector` and `PollingConnector`. Responsibility of this part of the code is to get flags from upstream and load it into core engine. I refactored loading part from `load_features` function into abstract `BaseConnector`. Other helper, `fetch_and_load`, exists now as a method in `PollingConnector`

## Type of change

- Refactoring

# How Has This Been Tested?

- [x] Unit tests
- [x] Spec Tests

# Checklist:


UnleashClient/connectors/base_connector.py
```diff
@@ -0,0 +1,63 @@
...
+                "Please make sure client can communicate with Unleash server!"
+            )
+            return
+
+        try:
+            payload = (
+                json.dumps(feature_provisioning)
+                if isinstance(feature_provisioning, dict)
+                else feature_provisioning
+            )
```
sighphyre, line 50: What's the need for coercing the payload here? We should always have a string, shouldn't we?

UnleashClient/__init__.py
```diff
@@ -185,7 +184,32 @@ def __init__(
...
+
+        # Bootstrapping
+        if self.unleash_bootstrapped:
+            OfflineConnector(
+                engine=self.engine,
+                cache=self.cache,
+                scheduler=None,
+            ).load_features()
+
+        self.connector: Union[OfflineConnector, PollingConnector] = None
```
sighphyre, line 205: We went through all that work of making the connector types swappable. Let's indicate that with the type hinting (you'll need to import this if you take the suggestion)

```suggestion
        self.connector: Union[BaseConnector] = None
```

sighphyre, line 205: Absolute nit but can we move this up to where all the other fields on this class are declared?

UnleashClient/__init__.py
```diff
@@ -185,7 +184,32 @@ def __init__(
...
+            self.engine.register_custom_strategies(custom_strategies)
+
+        self.strategy_mapping = {**custom_strategies}
+
+        # Client status
+        self.is_initialized = False
+
+        # Bootstrapping
+        if self.unleash_bootstrapped:
+            OfflineConnector(
```
sighphyre, line 199: Okay that's pretty elegant! I'm tempted to call this connector BootstrapConnector, just based on this. That's really what it's doing. Fine if you like OfflineConnector though

UnleashClient/__init__.py
```diff
@@ -304,47 +292,58 @@ def initialize_client(self, fetch_toggles: bool = True) -> None:
                     )
 
                 if fetch_toggles:
-                    fetch_headers = {
+                    self.unleash_scheduler.start()
```
sighphyre, line 295: I'm not sure this pattern of calling the scheduler start like this is going to work. That seems to raise an exception if it's already started. I see the check for the running state in the metrics but that looks like a hack to me and it's going to go wrong at some point. Can't we just call it once in this method like we did before?

UnleashClient/__init__.py
```diff
@@ -412,7 +412,8 @@ def destroy(self) -> None:
                 engine=self.engine,
             )
 
-        self.unleash_scheduler.shutdown()
+        if getattr(self.unleash_scheduler, "state", None) == STATE_RUNNING:
```
sighphyre, line 415: This feels wrong. I want to say the right way to do this is to try and shut down the scheduler and if you're worried then trap the SchedulerNotRunningError exeption

tests/unit_tests/connectors/test_offline_connector.py
```diff
@@ -0,0 +1,88 @@
...
+
+    connector = OfflineConnector(
+        engine=engine,
+        cache=temp_cache,
+        scheduler=scheduler,
+    )
+
+    connector.load_features()
+
+    assert not engine.is_enabled("testFlag", {})
```
sighphyre, line 40: This assert here tells us nothing, this will always pass since engine.is_enabled will return None here if there's no feature

sighphyre (commented):
Cool! Looking promising!

UnleashClient/__init__.py
Tymek: Maybe broad 'catch' would be better?

UnleashClient/__init__.py
Tymek: _init_scheduler? yes, but it triggers 'too many statements' linting error

PR #387, opened 2026-01-21
feat: impact metrics

kwasniew:
# Description

**Adds full impact metrics support to this SDK**

This PR is the first example of impact metrics implemented mostly in Yggdrasil, with the SDK doing the wrapper work.
The actual change is <100 LOC. Most of the PR is README and tests.

What the SDK has to do is:
* orchestrate calls to the Ygg engine 
* add label resolution for impact metrics
* verify everything works together in a test

**Interesting decisions:**
* Node restores regular Unleash bucket metrics on failure, Python doesn't. This is pre-existing behavior. I didn't want to change that. However, we have smart restoration for impact metrics that is identical to the Node SDK's impact metric restoration on failure. Please shout if you disagree with this decision :) 
* The ImpactMetrics class is a wrapper around the Ygg engine with extra label resolution added on top 
* For testing, I decided to use file bootstrap to simplify setup and avoid mocks. I also test using the Unleash Client API and check what was sent over the wire. This test gives high ROI without testing the internals. The only thing I didn't like—but I found similar examples in other tests—is how we trigger metric sending. The public API doesn't expose an explicit send metrics method, so I use the internal method for metric sending (aggregate_and_send_metrics). I still prefer it to an arbitrary wait for the scheduler.

# How Has This Been Tested?

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,58 @@ client.is_enabled("testFlag")
 
 ```
 
+### Impact metrics
+
+Impact metrics are lightweight, application-level time-series metrics stored and visualized directly inside Unleash. They allow you to connect specific application data, such as request counts, error rates, or latency, to your feature flags and release plans.
+
+These metrics help validate feature impact and automate release processes. For instance, you can monitor usage patterns or performance to determine if a feature meets its goals.
+
+The SDK automatically attaches context labels to metrics: `appName` and `environment`.
+
+#### Counters
+
+Use counters for cumulative values that only increase (total requests, errors):
+
+```python
+client.impact_metrics.define_counter(
+    "request_count",
+    "Total number of HTTP requests processed"
+)
+
+client.impact_metrics.increment_counter("request_count")
+```
+
+#### Gauges
+
+Use gauges for fluctuating values like memory usage or active connections:
+
+```python
+import psutil
+
+client.impact_metrics.define_gauge(
+    "memory_usage",
+    "Current memory usage in bytes"
+)
+
+current_memory = psutil.Process().memory_info().rss
+client.impact_metrics.update_gauge("memory_usage", current_memory)
+```
+
+#### Histograms
+
+Histograms measure value distribution (request duration, response size):
+
+```python
+client.impact_metrics.define_histogram(
+    "request_time_ms",
+    "Time taken to process a request in milliseconds"
+)
+
+client.impact_metrics.observe_histogram("request_time_ms", 125)
+```
+
+Impact metrics are batched and sent using the same interval as standard SDK metrics.
+
 ### Custom cache
 
 By default, the Python SDK stores feature flags in an on-disk cache using fcache. If you need a different storage backend, for example, Redis, memory-only, or a custom database, you can provide your own cache implementation.
--- a/README.md
+++ b/README.md
@@ -342,18 +342,15 @@ client.impact_metrics.increment_counter("request_count")
 
 #### Gauges
 
-Use gauges for fluctuating values like memory usage or active connections:
+Use gauges for point-in-time values that can go up or down:
 
 ```python
-import psutil
-
 client.impact_metrics.define_gauge(
-    "memory_usage",
-    "Current memory usage in bytes"
+    "total_users",
+    "Total number of registered users"
 )
 
-current_memory = psutil.Process().memory_info().rss
-client.impact_metrics.update_gauge("memory_usage", current_memory)
+client.impact_metrics.update_gauge("total_users", user_count)
 ```
 
 #### Histograms
--- a/README.md
+++ b/README.md
@@ -363,7 +363,8 @@ Histograms measure value distribution (request duration, response size):
 ```python
 client.impact_metrics.define_histogram(
     "request_time_ms",
-    "Time taken to process a request in milliseconds"
+    "Time taken to process a request in milliseconds",
+    [50, 100, 200, 500, 1000]
 )
 
 client.impact_metrics.observe_histogram("request_time_ms", 125)
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -284,7 +284,6 @@ class UnleashClient:
     def is_initialized(self):
         return self._run_state == _RunState.INITIALIZED
 
-
     def initialize_client(self, fetch_toggles: bool = True) -> None:
         """
         Initializes client and starts communication with central unleash server(s).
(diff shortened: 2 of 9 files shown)


UnleashClient/periodic_tasks/send_metrics.py
```diff
@@ -19,6 +19,7 @@ def aggregate_and_send_metrics(
     engine: UnleashEngine,
 ) -> None:
     metrics_bucket = engine.get_metrics()
+    impact_metrics = engine.collect_impact_metrics()
```
sighphyre: If I remember the binding code, this can potentially raise an exception. Think that's pretty unlikely but how do you feel about making this flow not bork metrics if this fails?

UnleashClient/__init__.py
```diff
@@ -205,6 +206,9 @@ def __init__(
         self.fl_job: Job = None
         self.metric_job: Job = None
         self.engine = UnleashEngine()
+        self._impact_metrics = ImpactMetrics(
+            self.engine, self.unleash_app_name, self.unleash_environment
```
sighphyre: We don't want to do what we do in Node and get this from the token? I'm going to deprecate this environment across the SDKs over the next few months, just a heads up

UnleashClient/__init__.py
```diff
@@ -273,6 +277,10 @@ def connection_id(self):
     def is_initialized(self):
         return self._run_state == _RunState.INITIALIZED
 
+    @property
+    def impact_metrics(self) -> ImpactMetrics:
```
sighphyre: Ah, found the Java developer! So I don't think this adds anything to our lives in the current shape. This is necessary pattern in Java but the magic of Python properties is that we can do this later if we ever feel the need without impacting the public API. I would just expose the field directly to be honest

README.md
```diff
@@ -319,6 +319,56 @@ client.is_enabled("testFlag")
 
 ```
 
+### Impact metrics
```
sighphyre, line 322: Thank you!

UnleashClient/impact_metrics.py
```diff
@@ -0,0 +1,84 @@
...
+    def observe_histogram(
+        self,
+        name: str,
+        value: float,
+        flag_context: Optional[MetricFlagContext] = None,
+    ) -> None:
+        labels = self._resolve_labels(flag_context)
+        self._engine.observe_histogram(name, value, labels)
+
+    def _resolve_labels(
```
sighphyre, line 74: I don't think this is wrong but it does look like "Java developer slipped, fell and landed in the Python runtime" to me. What you have is clear so I'm going to make some recommendations based on what I would do given my Python background and you can pick and choose some or none of this

I think the creation of the dict here is a bit unfortunate. We close over the environment and appName fields but literally only use them here to new this dict up. Smells like a way to work around Python's lack of destructuring. Good news, Python does have destructuring though!

This is also definitely begging to be a list comprehension with destructuring. That may or may not read okay to someone without a Python background but it reads fine to me

This is how I would have done this:
``` python
def __init__(self, engine: UnleashEngine, app_name: str, environment: str):
    self._engine = engine
    self._base_labels = {
        "appName": app_name,
        "environment": environment,
    }


def _variant_label(self, flag_name: str, ctx) -> str:
    variant = self._engine.get_variant(flag_name, ctx)
    if variant and variant.enabled:
        return variant.name
    if variant and variant.feature_enabled:
        return "enabled"
    return "disabled"


def _resolve_labels(
    self, flag_context: Optional[MetricFlagContext]
) -> Dict[str, str]:
    if not flag_context:
    	## Just a lil defensive copying so we don't leak mutable state
        return dict(self._base_labels)

    return {
        **self._base_labels,
        **{
            flag: self._variant_label(flag, flag_context.context)
            for flag in flag_context.flag_names
        },
    }

```

UnleashClient/impact_metrics.py
```diff
@@ -0,0 +1,84 @@
...
+
+from yggdrasil_engine.engine import UnleashEngine
+
+
+@dataclass
+class MetricFlagContext:
+    """Context for resolving feature flag values as metric labels."""
+
+    flag_names: List[str] = field(default_factory=list)
+    context: Dict[str, Any] = field(default_factory=dict)
```
sighphyre, line 12: Getting some mixed messages from this field. The class is a context but so is this. Is it contexts all the way down?

Weirdly enough because of this, I have no context for what this actually represents

sighphyre (approved):
Yeah cool, this shaped up really nicely

UnleashClient/__init__.py
kwasniew: That's a good point. I remember fixing it in Node so we can do the same here.

UnleashClient/__init__.py
kwasniew: Honestly I didn't want to do this but got inspired by  def connection_id(self) which should probably be a regular property

UnleashClient/impact_metrics.py
kwasniew: yup I like your version more. thanks for pointing it towards more idiomatic python

UnleashClient/impact_metrics.py
kwasniew: this is consistent with what we called it in Node SDK. This is our user context so that we can evaluate flag_names to enabled/disabled/variant_name with a user context. I agree it may be confusing here since in Node Context has a proper shape:
```ts
export interface Context {
  [key: string]: string | Date | undefined | number | Properties;
  currentTime?: Date;
  userId?: string;
  sessionId?: string;
  remoteAddress?: string;
  environment?: string;
  appName?: string;
  properties?: Properties;
}

```

kwasniew: I think this SDK uses plain dict for user context everywhere and I don't want to make a big change in this PR

UnleashClient/periodic_tasks/send_metrics.py
kwasniew: hmm, I though it should never happen. But to be safe with some weird internal errors I can wrap it with an error handler

UnleashClient/impact_metrics.py
sighphyre: Ah! It's the actual context. Yep, makes sense, let's keep it the same. Context needs a proper type some day, today is not that day

UnleashClient/__init__.py
sighphyre: It probably should. Dunno why it's like that haha, probably AI code that got a bit too excited

Issue #303, opened 2024-03-06
please ditch fcache as it does not work with multiprocessing on Windows


masariello:
**Describe the bug**

fcache has a known issue with multiprocessing (tsroten/fcache#26) that has been ignored tor 5 years.

Here's an example of the errors

```
2024-03-06 10:54:33.478 - ERROR - apscheduler.executors.unleash_executor_YTIN7P - run_job - Job "fetch_and_load_features (trigger: interval[0:01:00], next run at: 2024-03-06 10:55:32 CST)" raised an exception
Traceback (most recent call last):
  File "C:\Users\userName\AppData\Roaming\Python\Python39\site-packages\apscheduler\executors\base.py", line 125, in run_job
    retval = job.func(*job.args, **job.kwargs)
  File "C:\Users\dchuserNameoi\AppData\Roaming\Python\Python39\site-packages\UnleashClient\periodic_tasks\fetch_and_load.py", line 43, in fetch_and_load_features
    cache.set(ETAG, etag)
  File "C:\Users\userName\AppData\Roaming\Python\Python39\site-packages\UnleashClient\cache.py", line 129, in set
    self._cache[key] = value
  File "C:\Users\userName\AppData\Roaming\Python\Python39\site-packages\fcache\cache.py", line 267, in __setitem__
    self._write_to_file(filename, value)
  File "C:\Users\userName\AppData\Roaming\Python\Python39\site-packages\fcache\cache.py", line 254, in _write_to_file
    os.chmod(filename, self._mode)
OSError: [WinError 6800] The function attempted to use a name that is reserved for use by another transaction: 'C:\\Users\\userName\\AppData\\Local\\PROD\\PROD\\Cache\\cache\\65746167'
```

**To Reproduce**

Please see StackOverflow linked in tsroten/fcache#26

sighphyre:
Hey @masariello, this probably isn't something that'll get resolved in the next few weeks. But we do have tentative plans to rework the caching anyway due to an upcoming project on this SDK

Is this actually affecting you, outside of causing trouble in the logs? Writing to backup, shouldn't be a process that needs to happen to ensure the SDK works correctly

masariello:
Yes. I hit the problem reproduced in the linked fcache issues

calum-pledge-io:
As an extra motivation, fcache also isn't compatible with many serverless environments out-of-the-box.
[...]

PR #391, opened 2026-02-13
fix: do not delete cache on destroy

### LINKED ISSUE #390: UnleashClient.destroy() removes shared cache directory across processes
**Describe the bug**
The default Unleash cache implementation (`FileCache`, based of `fcache.cache.FileCache`) uses `platformdirs.user_cache_dir()` to compute the cache directory. On Linux systems, this resolves to `${HOME}/.cache/APP_NAME/cache` (where `APP_NAME` is defined via `UnleashClient`).

When `UnleashClient.destroy()` is executed (i.e., to ensure proper metrics flush before shutdown), the **entire cache directory is removed** as part of the cleanup process. As a result, every other running process **using the same `APP_NAME`** (and, therefore, relying on the **same cache path**) fails when attempting to use the cache.

**To Reproduce**
1. Initialize `UnleashClient` and register `client.destroy()` in an `atexit.register` handler.
2. Start a long running process (i.e., a FastAPI app).
3. Open an interactive shell to the FastAPI app in a new process.
4. Close the interactive shell. It will trigger `UnleashClient.destroy()` as part of the shutdown.
5. `PollingConnector._fetch_and_load()` will start raising errors when writing to the long running process' cache.

**Expected behavior**
We would expect both the FastAPI process and the interactive shell to use its own, isolated cache dir, but the cache dir is currently shared between every process running with the same `APP_NAME`.

**Logs**
```python
Traceback (most recent call last):
  File "shutil.py", line 856, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpjoiroou9' -> '/root/.cache/my-project-name/cache/2f636c69656e742f6665617475726573'

Traceback (most recent call last):
  File "apscheduler/executors/base.py", line 131, in run_job
    retval = job.func(*job.args, **job.kwargs)
  File "UnleashClient/connectors/polling_connector.py", line 72, in _fetch_and_load
    self.cache.set(FEATURES_URL, state)
  File "UnleashClient/cache.py", line 129, in set
    self._cache[key] = value
  File "fcache/cache.py", line 274, in __setitem__
    self._write_to_file(filename, value)
  File "fcache/cache.py", line 259, in _write_to_file
    rename(tmp, filename)
  File "shutil.py", line 876, in move
    copy_function(src, real_dst)
  File "shutil.py", line 468, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "shutil.py", line 262, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/my-project-name/cache/2f636c69656e742f6665617475726573'
```

**Additional context**
- **Python**: 3.13.7
- **UnleashClient**: 6.5.1
- **System**: Debian Trixie (`python:3.13.7-slim-trixie`)
- **Architecture**: ARM64

gastonfournier:
# Description

Fix `UnleashClient.destroy()` behavior so it no longer deletes the default SDK FileCache on shutdown.

Previously, shutdown always called `cache.destroy()`. For the default FileCache, this could remove shared on-disk cache data and break other live processes using the same app_name cache path (reported in #390). The fix preserves shutdown behavior for connector/scheduler/metrics while skipping cache deletion for default FileCache. Custom cache implementations still receive destroy() as before.

This is a targeted mitigation for the teardown race in #390. It does not fully address broader multiprocessing limitations of fcache discussed in #303.

Fixes #390

## Type of change

# How Has This Been Tested?
Added/updated unit tests:
- test_destroy_skips_default_file_cache_destroy
- test_destroy_calls_custom_cache_destroy

Also updated cache fixture isolation in unit tests (tmpdir.strpath) to avoid cross-test cache leakage now that default FileCache is not deleted during client shutdown.

- [x] Unit tests
- [ ] Spec Tests

# Checklist:

### DIFF
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -487,7 +487,6 @@ class UnleashClient:
             except Exception as exc:
                 LOGGER.warning("Exception during scheduler teardown: %s", exc)
 
-
     @staticmethod
     def _get_fallback_value(
         fallback_function: Callable, feature_name: str, context: dict
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -487,10 +487,14 @@ class UnleashClient:
             except Exception as exc:
                 LOGGER.warning("Exception during scheduler teardown: %s", exc)
 
-            try:
-                self.cache.destroy()
-            except Exception as exc:
-                LOGGER.warning("Exception during cache teardown: %s", exc)
+            # The default SDK FileCache is disk-backed and can be shared by
+            # multiple processes for the same app name. Avoid deleting it
+            # during shutdown to prevent cross-process cache races.
+            if not isinstance(self.cache, FileCache):
+                try:
+                    self.cache.destroy()
+                except Exception as exc:
+                    LOGGER.warning("Exception during cache teardown: %s", exc)
 
     @staticmethod
     def _get_fallback_value(
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -487,14 +487,6 @@ class UnleashClient:
             except Exception as exc:
                 LOGGER.warning("Exception during scheduler teardown: %s", exc)
 
-            # The default SDK FileCache is disk-backed and can be shared by
-            # multiple processes for the same app name. Avoid deleting it
-            # during shutdown to prevent cross-process cache races.
-            if not isinstance(self.cache, FileCache):
-                try:
-                    self.cache.destroy()
-                except Exception as exc:
-                    LOGGER.warning("Exception during cache teardown: %s", exc)
 
     @staticmethod
     def _get_fallback_value(
--- a/UnleashClient/__init__.py
+++ b/UnleashClient/__init__.py
@@ -487,6 +487,14 @@ class UnleashClient:
             except Exception as exc:
                 LOGGER.warning("Exception during scheduler teardown: %s", exc)
 
+            # Disk-backed FileCache instances can be shared across processes.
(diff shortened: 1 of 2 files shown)


UnleashClient/__init__.py
```diff
@@ -487,10 +487,14 @@ def destroy(self) -> None:
             except Exception as exc:
                 LOGGER.warning("Exception during scheduler teardown: %s", exc)
 
-            try:
```
sighphyre, line 490: I think we can just throw this out. There's never a reason to destroy the cache

sighphyre (approved):
I think we can just remove the cache destruction code here but this is probably safe

UnleashClient/__init__.py
```diff
@@ -487,10 +487,14 @@ def destroy(self) -> None:
...
-            except Exception as exc:
-                LOGGER.warning("Exception during cache teardown: %s", exc)
+            # The default SDK FileCache is disk-backed and can be shared by
+            # multiple processes for the same app name. Avoid deleting it
+            # during shutdown to prevent cross-process cache races.
+            if not isinstance(self.cache, FileCache):
+                try:
+                    self.cache.destroy()
+                except Exception as exc:
+                    LOGGER.warning("Exception during cache teardown: %s", exc)
```
gastonfournier: There's no reason to delete a cache
```suggestion
```

gastonfournier: Related to https://github.com/Unleash/unleash-python-sdk/pull/391/changes#r2803309752

PR #395, opened 2026-07-17
feat: add sdk flavor metadata

irstavr:
Adds skd flavor metadata when building up and sets them properly when register and send metrics.
Get it ready for use in the provider side

### DIFF
--- a/UnleashClient/utils.py
+++ b/UnleashClient/utils.py
@@ -9,6 +9,17 @@ from requests import Response
 LOGGER = logging.getLogger("UnleashClient")
 
 
+def sdk_flavor_fields(
+    sdk_flavor: Any = None, sdk_flavor_version: Any = None
+) -> dict:
+    fields = {}
+    if sdk_flavor:
+        fields["sdkFlavor"] = sdk_flavor
+    if sdk_flavor_version:
+        fields["sdkFlavorVersion"] = sdk_flavor_version
+    return fields
+
+
 class InstanceAllowType(Enum):
     BLOCK = 1
     WARN = 2
--- a/UnleashClient/utils.py
+++ b/UnleashClient/utils.py
@@ -9,15 +9,6 @@ from requests import Response
 LOGGER = logging.getLogger("UnleashClient")
 
 
-def sdk_flavor_fields(sdk_flavor: Any = None, sdk_flavor_version: Any = None) -> dict:
-    fields = {}
-    if sdk_flavor:
-        fields["sdkFlavor"] = sdk_flavor
-    if sdk_flavor_version:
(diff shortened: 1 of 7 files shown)


UnleashClient/utils.py
```diff
@@ -9,6 +9,15 @@
 LOGGER = logging.getLogger("UnleashClient")
 
 
+def sdk_flavor_fields(sdk_flavor: Any = None, sdk_flavor_version: Any = None) -> dict:
+    fields = {}
+    if sdk_flavor:
+        fields["sdkFlavor"] = sdk_flavor
+    if sdk_flavor_version:
+        fields["sdkFlavorVersion"] = sdk_flavor_version
+    return fields
```
sighphyre: A few small thoughts here:

1) Do we need this? We use this twice and it looks like it's only done to enforce data types on the registration/metrics. My go to rule around abstractions is usually when I hit 3 callsites or above
2) If you really want to keep it, I'm okay with that, but let's clean up the type checking, right now we're doing Any, I think we should be able to get away with Optional[str] instead here
3) If you really want the abstraction but want to do it in a more Python fashion, I would recommend converting the parameters in the publicly facing constructor into a small data class and passing that around and then de-structuring that into the metrics/registration payloads right at the end of the call chain. Something like this

``` python
@dataclass
class Flavour:
    flavour_type: str
    flavour_version: str

    def to_dict(self) -> dict[str, str]:
        return {
            "flavourType": self.flavour_type,
            "flavourVersion": self.flavour_version,
        }
```

Then you can use it like


```python
   {
       #some other fields
        **asdict(flavour_class)
   }
```

sighphyre (approved):
Seems sane to me. Left a suggestion around cleaning up the way we pass around the flavour fields but it's not a biggie, up to you

UnleashClient/utils.py
irstavr: Fair enough with the 3 :) thanks for the thorough read. Refactored it following your comments.

I went with # 1 - without abstraction. As both 2 callsites just enforced the wire shape. I removed the helper out and refactored it for building the payload inline - in a pair of "if SDK_FLAG" guards at each payload builder. 

Thisˆ takes care of #2 as well - the `any` is gone, and the params are `Optional[str]` now. I kept the "omit when unset" behaviour, so a plain SDK client's payload doesn't change and we never emit `sdkFlavor: null`

#3: I really liked the flavour dataclass and I'll happily use it as soon as this grows. I'm not sure about adding the class and the public-constructor change for two optional strings across two sites - as you said. I think it might be too soon to do that.

sighphyre (approved):
Yep, come together nicely

PR #397, opened 2026-07-24
fix: normalize metrics, features and registration urls

sighphyre:
Fixes an issue in the URL generation code for metrics, registration and features by preventing double slashes in the URL path when the user sets a Unleash server path with a trailing slash

### DIFF
--- a/UnleashClient/api/urls.py
+++ b/UnleashClient/api/urls.py
@@ -1,0 +1,2 @@
+def get_url(url: str, path: str) -> str:
+    return f"{url.rstrip('/')}{path}"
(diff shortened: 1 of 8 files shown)


gastonfournier (approved):
LGTM!

UnleashClient/api/urls.py
```diff
@@ -0,0 +1,2 @@
+def get_url(url: str, path: str) -> str:
```
ppicom: **nitpick**[^ref]: Have we considered a more specific name (e.g. `normalize_trailing_slash`)

`get` suggests me either a getter from a class or a function that will fetch something over the network

[^ref]: https://conventionalcomments.org/

ppicom (approved):

PR #398, opened 2026-07-24
chore: attempt to force coveralls status

sighphyre:
Posts the legacy coverage/coveralls commit status from CI after the Coveralls upload completes.

Coveralls is receiving coverage, but GitHub branch protection is still waiting forever for that required status context. This keeps the existing upload and explicitly reports success/failure for the required check.

Bit of a dirty hack but I just want coveralls to stop blocking my life


irstavr (approved):
yeee!

Issue #240, opened 2023-03-15
async support


raphaelauv:
Async support based on httpx or aiohttp would be great

thanks all

daveleek:
Thank you for the suggestion @raphaelauv, I'll bring this up!

daveleek:
Hi again @raphaelauv! Thank you for the suggestion, it is something we'd like to look into in the future, and if you have a PR then we'd love to take a look at that!

snosratiershad:
Dear @daveleek, I think the async support is essential but it's not backward compatible and if we can do it in a backward compatible mode, it would not benefit from asynchronous support. I think the better idea is to develop a dedicated module inside or externally. that uses some functionalities but provides async interfaces and processes. there's a lot of related examples in Python libraries such as database integration libs, 3th-party API clients, etc. How do you think about it?

snosratiershad:
Dear @philipbjorge, there is only one thread in SDK (even if we support async in the future, it's still one thread and a single-thread event loop).  are you trying to use unleash-client-python via different processes in a single application? or I'm missing some info

philipbjorge:
@snosratiershad --
We're trying to use this client in a FastAPI application.
I want to make sure that calling `is_enabled(...)` isn't going to block our main thread with things like blocking network calls.

yjabri:
I'm also interested in async support. In our synchronous apps we currently make use of a custom `event_callback` that publishes messages for analytics. It'd be awesome if we could use `async` event callbacks!

I think the `httpx` project utilizes a good strategy for organizing both sync and async clients. Both the [sync client](https://github.com/encode/httpx/blob/master/httpx/_client.py#L569) and [async client](https://github.com/encode/httpx/blob/master/httpx/_client.py#L1293) subclass a common [base client](https://github.com/encode/httpx/blob/master/httpx/_client.py#L158). I think there would be enough shared logic that such an approach could potentially be fruitful here.

@snosratiershad, would such an approach be welcomed in this project?

yjabri:
After spending a little more time getting more acquainted with the code, one challenge I'm not sure how to resolve is doing non-blocking IO given the dependency on `fcache`.

Unless I'm missing something I imagine the primary use case for an async client would be to call `is_enabled` and `get_variant` asynchronously. One could probably get away with synchronous client initialization and I think that would drastically simplify an `async`-ish implementation.

I wonder if any other projects do anything like that.

snosratiershad:
[...]

PR #399, opened 2026-07-27
refactor: new /clients directory

ppicom:
## Summary

Making way for what's to come (yak shaving a bit), this PR includes two tiny rearrangements:

- **refactor: move `extract_environment_from_headers` to utils.py**

This one is highly based on my opinion. The environment function seems to fit quite well inside of utils.py and that would be one less file to maintain.

- **refactor: create a new `/clients` directory**

This one prepares the directory structure for the implementation of #240, mostly to avoid conflict hells further down the line and to not mix more than one change in the same PR.


tests/unit_tests/api/test_feature.py
```diff
@@ -49,7 +49,7 @@ def test_get_feature_toggle(response, status, calls, expected):
         headers={"etag": ETAG_VALUE},
     )
 
-    (result, etag) = get_feature_toggles(
```
ppicom: Somehow, there were linting errors on `main`

UnleashClient/connectors/polling_connector.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         self.job = None
 
     def _fetch_and_load(self):
-        (state, etag) = get_feature_toggles(
```
ppicom: Same. Preexisting linting errors from the main branch

irstavr (commented):
There are 3 files here that only relate to linting issues. We could split them up in a different PR to help with the review?!

ppicom:
Sure! Let me chedrry pick them

ppicom:
> There are 3 files here that only relate to linting issues. We could split them up in a different PR to help with the review?!

I plan on merging the linting errors in this PR https://github.com/Unleash/unleash-python-sdk/pull/400 before I remove them from here. Sounds good?

sighphyre (approved):
Failing check on the API token isn't new so I don't think that's a problem with this PR - don't think Github's suggestion is a good one though, this can probably either take a hash of the token or trumcate is to 6 characters after the project, so turn something like production-env:default-project:a7d651bae8aaefdb9179 -> production-env:default-project:a7d651, enough to not make it useful but still enough to make it recognizable. Can chat on Slack if you need  

I very much don't agree with collapsing the environment_resolver into utils, because I think utils is where good code structure goes to die. That being said, this is a taste thing and not a blocker.

Structure is nice, change makes sense, looks good! Happy if you merge this

PR #400, opened 2026-07-27
refactor: fix linting issues from the main branch

ppicom:
The goal of this merge request is to fix the linting issues that were on main so that, once merged, I can pick them out of https://github.com/Unleash/unleash-python-sdk/pull/399


irstavr (approved):
thanks for cleaning up 🧹

PR #401, opened 2026-07-27
fix: make environment and project visible in log line

ppicom:
These changes come after #400. Failed to include these changes when amending the previous merge request.

The goal of this changes is for the obfuscation to only be applied to the part of the API key that corresponds to _the actual cryptographic key_


sighphyre (approved):
Ye!

PR #403, opened 2026-08-04
Create new EventDispatcher to move user-provided callback execution out of the main thread

ppicom:
## Summary

The pull request introduces the implementation of `EventDispatcher`. Its goal is to move all event-and-callback-related code to the same file. At the same time, the `EventDispatcher` is based on threads: the "callers" of the dispatcher will use `emit_event` to queue an event, and a worker thread managed by the dispatcher will drain the queue.

In this pull request, the Dispatcher is not yet wired to anyone.

## Reason for these changes

This is the initial step towards solving two problems:
1. the user-provided event callback currently runs in the hot path, making it possible for users to stall `is_enabled` indefinitely; and
2. the async counterpart to the current `UnleashClient` (i.e. `AsyncUnleashClient`) will also support sending events to a callback

By introducing a single class that deals with event-sending as a whole, code will be shared and we ensure we don't run it in the hot path anymore (as long as we publish events through the `EventDispatcher`).

### Threads and locks

`EventDispatcher` deals with locks and threads, which is always fun. Variables like `_closed`, as well as the two centinels ( `_SHUTDOWN` and `_FlushMarker`) help manage the synchronization of operations:

- To avoid race conditions
- To avoid leaving the Dispatcher in an inconsistent state (e.g. allowing to send after having invoked `close()`

### Backpressure

I discarded the option of adding a timeout or a number of retries to a user's callback. I'm happy to revisit the decision.

### DIFF
--- a/UnleashClient/events.py
+++ b/UnleashClient/events.py
@@ -149,7 +149,7 @@ class EventDispatcher:
     def close(self, timeout: float = DEFAULT_TIMEOUT) -> None:
         """
         close() signals the dispatcher to stop. It enqueues a Shutdown sentinel to the queue, and waits
-        for its signal to be set. The worker thread will eventually get to the sentinal, set its ``done``
+        for its signal to be set. The worker thread will eventually get to the sentinel, set its ``done``
         signal, which will wake this thread and allow it to join the worker thread.
 
         Even if the Shutdown sentinel cannot be enqueued, the worker thread will still be stopped and the dispatcher marked as closed.
--- a/UnleashClient/events.py
+++ b/UnleashClient/events.py
@@ -1,3 +1,6 @@
+from threading import Event
+
+
 import queue
 import threading
 import time
@@ -72,19 +75,18 @@ class _Shutdown:
     Sentinel that tells the worker to stop draining and exit.
     """
 
-    __slots__ = ()
+    def __init__(self):
+        self.done: Event = threading.Event()
 
     def __repr__(self) -> str:
         return "<_SHUTDOWN>"
-
-
-_SHUTDOWN = _Shutdown()
 
 
 _QueueItem = Union[BaseEvent, _Shutdown]
 
 DEFAULT_MAX_QUEUE_SIZE = 100
 DEFAULT_TIMEOUT = 2.0
+DEFAULT_PUT_TIMEOUT = 0.1
 
 
 class EventDispatcher:
@@ -125,7 +127,9 @@ class EventDispatcher:
         with self._lock:
             return self._dropped
 
-    def emit_event(self, event: BaseEvent) -> None:
+    def emit_event(
+        self, event: BaseEvent, timeout: float = DEFAULT_PUT_TIMEOUT
+    ) -> None:
         """
         Queues an event for delivery.  Never blocks the caller.  READY events after
         the first one are ignored.
@@ -138,27 +142,12 @@ class EventDispatcher:
             if self._closed:
                 return
 
-            if event.event_type is UnleashEventType.READY:
-                if self._ready_fired:
-                    return
-                self._ready_fired = True
-
-            self._start_worker()
-
             try:
-                self._queue.put_nowait(event)
-                return
+                self._start_worker()
+                self._enqueue_event(event, timeout=timeout)
             except queue.Full:
-                self._dropped += 1
-                should_warn = not self._warned_about_full_queue
-                self._warned_about_full_queue = True
-
-        if should_warn:
-            LOGGER.warning(
-                "Unleash event queue is full, events are being dropped. This usually means the event callback is too slow."
-            )
-        else:
-            LOGGER.debug("Event was dropped because queue is full.")
+                self._count_dropped_event()
+                self._maybe_warn_about_full_queue()
 
     def close(self, timeout: float = DEFAULT_TIMEOUT) -> None:
         """
@@ -169,19 +158,34 @@ class EventDispatcher:
             if self._closed:
                 return
             self._closed = True
-            thread = self._thread
 
-        if thread is None:
+            start_time = time.monotonic()
+            try:
+                shutdown_signal = self._signal_shutdown(
+                    timeout=_remaining(start_time, timeout)
+                )
+                _ = shutdown_signal.done.wait(timeout=_remaining(start_time, timeout))
+            except queue.Full:
+                pass
+            finally:
+                self._stop_worker(timeout=_remaining(start_time, timeout))
+
+    def _signal_shutdown(self, timeout: float) -> _Shutdown:
+        """Signals the worker to stop draining and exit.  Caller must hold ``self._lock``."""
+        signal = _Shutdown()
+        self._enqueue_event(signal, timeout=timeout)
+        return signal
+
+    def _stop_worker(self, timeout: float) -> None:
+        """Stops the worker thread.  Caller must hold ``self._lock``."""
+        if self._thread is None:
             return
 
-        deadline = time.monotonic() + timeout
-        try:
-            self._queue.put(_SHUTDOWN, timeout=timeout)
-        except queue.Full:
-            LOGGER.debug("Timed out signalling shutdown to the event dispatcher.")
+        if not self._thread.is_alive():
+            self._thread = None
             return
 
-        thread.join(max(0.0, deadline - time.monotonic()))
+        return self._thread.join(timeout=timeout)
 
     def _start_worker(self) -> None:
         """Starts a worker if one is not running.  Caller must hold ``self._lock``."""
@@ -193,6 +197,28 @@ class EventDispatcher:
         )
         self._thread.start()
 
+    def _enqueue_event(
+        self, event: Union[BaseEvent, _Shutdown], timeout: float
+    ) -> None:
+        if (
+            isinstance(event, UnleashEvent)
+            and event.event_type is UnleashEventType.READY
+        ):
+            if self._ready_fired:
+                return
+            self._ready_fired = True
+
+        self._queue.put(item=event, timeout=timeout)
+
+    def _count_dropped_event(self) -> None:
+        self._dropped += 1
+
+    def _maybe_warn_about_full_queue(self) -> None:
+        if not self._warned_about_full_queue:
+            LOGGER.warning(
+                "Unleash event queue is full, events are being dropped. This usually means the event callback is too slow."
+            )
+
     def _run(self) -> None:
         """_run is the inifinite loop that drains the queue of events.
 
@@ -201,9 +227,15 @@ class EventDispatcher:
             item = self._queue.get()
 
             if isinstance(item, _Shutdown):
+                item.done.set()
                 return
 
             try:
                 self._callback(item)
             except Exception as exc:
                 LOGGER.warning("Error in event callback: %s", exc)
+
+
+def _remaining(start_time: float, timeout: float) -> float:
+    """Returns the remaining time until the timeout expires."""
+    return max(0.0, timeout - (time.monotonic() - start_time))
--- a/UnleashClient/events.py
+++ b/UnleashClient/events.py
@@ -222,15 +222,17 @@ class EventDispatcher:
         The caller will wait for at most ``timeout`` seconds to enqueue the event.
         Caller must hold ``self._lock``.
         """
-        if (
-            isinstance(event, UnleashEvent)
-            and event.event_type is UnleashEventType.READY
-        ):
-            if self._ready_fired:
-                return
-            self._ready_fired = True
+        is_ready_event = (
+            isinstance(event, BaseEvent) and event.event_type is UnleashEventType.READY
+        )
+
+        if is_ready_event and self._ready_fired:
+            return
 
         self._queue.put(item=event, timeout=timeout)
+
+        if is_ready_event:
+            self._ready_fired = True
 
     def _count_dropped_event(self) -> None:
         self._dropped += 1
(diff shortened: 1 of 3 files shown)


UnleashClient/events.py
```diff
@@ -60,3 +65,168 @@ def features(self) -> dict:
         if not hasattr(self, "_parsed_payload"):
             self._parsed_payload = loads(self.raw_features)["features"]
         return self._parsed_payload
+
+
+_SHUTDOWN = object()
```
ppicom: On second thought, this sentinel is not super very idiomatic.

ppicom: Replaced by a class-based one.

UnleashClient/events.py
```diff
@@ -60,3 +65,181 @@ def features(self) -> dict:
...
+    def __repr__(self) -> str:
+        return "<_SHUTDOWN>"
+
+
+_SHUTDOWN = _Shutdown()
+
+
+class _FlushMarker:
+    """
+    Sentinel that lets a caller wait until everything queued ahead of it has been delivered.
```
gastonfournier: How does this work? When we start flushing, does this stop receiving new events?

UnleashClient/events.py
```diff
@@ -60,3 +65,181 @@ def features(self) -> dict:
...
+        with self._lock:
+            if self._closed:
+                return
+
+            if event.event_type is UnleashEventType.READY:
+                if self._ready_fired:
+                    return
+                self._ready_fired = True
+
+        self._enqueue(event)
```
gastonfournier: emit_event() checks _closed under _lock, releases it, and only then calls _enqueue(). close() can set _closed and put _SHUTDOWN between those steps. The event can then be placed after _SHUTDOWN, where the worker will never process it.

gastonfournier (commented):
Something suggested by gpt which I couldn't validate, sorry I didn't have time to do this properly 🙏 

> timeout makes close() return promptly, but it does not make shutdown successful. When inserting _SHUTDOWN times out, _closed prevents a later retry while the worker remains alive. Could we distinguish "closing requested" from "closed," or otherwise ensure the shutdown signal is eventually enqueued? The current docstring says "Stops the dispatcher," which is not true on this path.

UnleashClient/events.py
ppicom: Oh, that is a nice catch. I should probably enqueue while still holding the lock, yeah

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+
+    __slots__ = ("done",)
+
+    def __init__(self) -> None:
+        self.done = threading.Event()
+
+
+_QueueItem = Union[BaseEvent, _Shutdown, _FlushMarker]
+
+DEFAULT_MAX_QUEUE_SIZE = 10_000
```
sighphyre: 10k is wildly ambitious in my opinion. It's best to think about queue size as a latency smoothing tool. 

A queue under no load is empty virtually all the time. A queue that's connecting to a downstream that cannot handle the load is always full. A queue that's smoothing out hiccups is full in bursts. If you make your queue too large then the risk is that you're accounting for cases that cannot be handled gracefully but must perpetually choke on 10k items, which has knock of effects on memory.

I think 10k is beyond reasonable back pressure... what about 100?

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+            if event.event_type is UnleashEventType.READY:
+                if self._ready_fired:
+                    return
+                self._ready_fired = True
+
+            self._start_worker()
+
+            try:
+                self._queue.put_nowait(event)
+                return
```
sighphyre: I don't... think... this is a problem. But I wanna point out what I'm seeing and then someone can tell me they agree it's not a problem

Ready isn't necessarily sent because _queue.put_no_wait might decide to drop it. That should probably only happen under high event load, so you'd have to have one end user thread waiting for ready and another already poking doing a bunch of things that are sending events. Don't see why someone would do that, but gosh users do the strangest things sometimes

...and if they do, that second thread will block forever waiting for a READY event that never arrives (which is kinda making me feel terribly sad for that thread)

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+                self._warned_about_full_queue = True
+
+        if should_warn:
+            LOGGER.warning(
+                "Unleash event queue is full, events are being dropped. This usually means the event callback is too slow."
+            )
+        else:
+            LOGGER.debug("Event was dropped because queue is full.")
+
+    def flush(self, timeout: float = DEFAULT_TIMEOUT) -> bool:
```
sighphyre: Flush isn't locked so it races. There's a few subtle things there around what waiting for done means here because of that

...uh... but also... why do we need this? Not clear to me from this PR

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+
+        if thread is None:
+            return
+
+        deadline = time.monotonic() + timeout
+        try:
+            self._queue.put(_SHUTDOWN, timeout=timeout)
+        except queue.Full:
+            LOGGER.debug("Timed out signalling shutdown to the event dispatcher.")
+            return
```
sighphyre: Now what?

self._closed is marked as True so we can't re-enter, we didn't put the event on the queue so we won't actually stop. And we can never put any more events on the queue because self._closed is marked as True. Think that means we just spin forever, with no events, waiting for nothing. Alone. In sadness 😢

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+            self._queue.put(marker, timeout=timeout)
+        except queue.Full:
+            return False
+
+        return marker.done.wait(max(0.0, deadline - time.monotonic()))
+
+    def close(self, timeout: float = DEFAULT_TIMEOUT) -> None:
+        """
+        Stops the dispatcher.  Events already queued are delivered first, but only
+        for as long as the timeout allows.  Calling this more than once is a no-op.
```
sighphyre: I think this is subtly untrue. Our timeout here bounds how long this function will block for. But a return here doesn't actually mean the dispatcher was stopped, the thread is quietly getting on with its life, unjoined, free from the tyranny of the synchronization primitives without a way to stop it. Maybe that doesn't matter, dunno

UnleashClient/events.py
```diff
@@ -60,3 +65,195 @@ def features(self) -> dict:
...
+                continue
+
+            if isinstance(item, _Shutdown):
+                self._release_flush_markers()
+                return
+
+            try:
+                self._callback(item)
+            except Exception as exc:
+                LOGGER.warning("Error in event callback: %s", exc)
```
sighphyre: I can never freaking remember how Python logging works. Does this nom the traceback of this exception? I think it might. Do we care? Dunno, could be useful for users

sighphyre (commented):
Off the top of my head - think this should work fine for the sync client. Less certain about the async variant. Maybe the plan is to do something else there.

This is definitely, definitely going to cause hell when forked, so whether this works or not is going to depend on when and where the background thread is started

UnleashClient/events.py
ppicom: nope, not exactly. the idea is that `flush()` enqueues a marker and waits for it to be dequeued by the worker. But it does not close the event dispatcher.

UnleashClient/events.py
ppicom: I see what you're saying, it totally makes sense. Just to accumulate some more knowledge, when you say

> the risk is that you're accounting for cases that cannot be handled gracefully but must perpetually choke on 10k items

what do you mean?

ppicom:
```suggestion
DEFAULT_MAX_QUEUE_SIZE = 100
```

UnleashClient/events.py
ppicom: > ...uh... but also... why do we need this? Not clear to me from this PR

Super fair observation 🤔 `flush()` is not _needed_. It makes testing easier, though, because tests can "flush" the queue when they need to make assertions. Writing production code that is used just for testing is considered a smell, and now that I've had to explain it in this comment it makes a little less sense than it made yesterday!

> Flush isn't locked so it races. There's a few subtle things there around what waiting for done means here because of that

I for sure missed something, let me go over the code.

UnleashClient/events.py
ppicom: oh, and if this leaves a thread run wild, it sounds a bit like memory leak? Let me see.

UnleashClient/events.py
ppicom: Oh man, so true. This is so helpful, Simon! Let me see, because that does not sound very well.

UnleashClient/events.py
ppicom: uhmmm I'm missing the `exc_info=True` bit, apparently!

ppicom (commented):
Absolutely awesome review, thanks for that guys! Let me go back to the drawing board with your notes, and get back to you

UnleashClient/events.py
ppicom: My gut feeling tells me to switch from `put_nowait` to `put` with a timeout. Then, capture the exception and retry a couple of times. If all that fails, then we bubble up an exception.

I'm still thinking if, when all that happens, we then join the threads.

ppicom:
Following @sighphyre and @gastonfournier 's reviews, I've made a few changes to the pull request to make it simpler (I think).

## Bye bye, `flush()`

I thought it made sense such a method to ease the testing. But nope. I've changed my mind after what Simon pointed out. The public interface of the EventDispatcher now is only made of `emit_event()` and `close()`.

Therefore, there is no `_FlushMarker` and none of that nonsense.

Testing is achieved through the public API and some "dedicated" callbacks that are part of the helpers of the test suite (e.g. `RecordingCallback`, `BlockingCallback`, etc)

## Unjoined thread

My previous changes left the thread unjoined regardless of whether `close()` worked or not. Shame! Now, the `finally` block is a catch-all that will join the thread and leave the EventDispatcher in a _closed state.

If all goes well, then good. If not, at least there is no dangling thread.

## Shutting down

The shutdown of the event dispatcher is achieved by queueing a sentinel. `close()` will block until the worker picks the sentinel up and sets its signal (waking up the parent thread). When that happens, it means that the worker has drained the queue, and it is safe to join the thread and be done with it.

## `__del__` and the garbage collector

I thought it made sense to also join the thread in the destructor of a class in case a user could have left a reference to the event dispatcher. But I have discarded the idea to not entangle this further. I'm not sure it brings much benefits.

UnleashClient/events.py
```diff
@@ -60,3 +66,193 @@ def features(self) -> dict:
...
+        self._queue: queue.Queue[_QueueItem] = queue.Queue(maxsize=max_size)
+        self._lock = threading.Lock()
+        self._thread: Optional[threading.Thread] = None
+        self._dropped = 0
+        self._warned_about_full_queue = False
+        self._ready_fired = False
+        self._closed = False
+
+    def __del__(self):
+        self.close()
```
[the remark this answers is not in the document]
ppicom: Deleted the destructor.

UnleashClient/events.py
```diff
@@ -60,3 +66,193 @@ def features(self) -> dict:
...
+                    )
+                    _ = shutdown_signal.done.wait(
+                        timeout=_remaining(start_time, timeout)
+                    )
+            except queue.Full:
+                # Even if Shutdown could not be queued, we'll still stop the worker thread and mark as ``_closed``.
+                # Not much we can do about it, but at least we won't leave the thread running.
+                pass
+            finally:
+                self._stop_worker(timeout=_remaining(start_time, timeout))
```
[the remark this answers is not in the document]
ppicom: I can see how this makes sense and how it can affect correctness a bit. I'm not sure how often users will reference the Dispatcher directly, and close it in their own callback.

But I've applied the suggestion so we don't allow a thread to close itself (which raises a `RuntimeError`)

UnleashClient/events.py
```diff
@@ -60,3 +66,193 @@ def features(self) -> dict:
...
+        """
+        if (
+            isinstance(event, UnleashEvent)
+            and event.event_type is UnleashEventType.READY
+        ):
+            if self._ready_fired:
+                return
+            self._ready_fired = True
+
+        self._queue.put(item=event, timeout=timeout)
```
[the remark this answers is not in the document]
ppicom: Makes sense, but did not apply the suggestion as-is.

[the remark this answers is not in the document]
ppicom: Fixed in the last commit

UnleashClient/events.py
sighphyre: So imagine a system under load. That system is not happy with life and it's sending events faster than downstream can process them. An event is at a thumbsuck about 3-5kb, some of them are fatbois are are probably closer to 100kb. So with a 10000 item backbuffer you have roughly 45mb just hanging around doing nothing for your life. System is already miserable so adding to its memory pressure just makes it more miserable. 100 items in your queue would mean about 450kb hanging around, which is much friendlier to a miserable system. It does mean dropping events sooner, but honestly, if you're going beyond 100 items that's going to happen sooner or later unless you have some weird and insane latency bursts

gastonfournier: Thinking aloud here:
- Should we bulkhead these events? I can see some are more important than others but this would create another race condition and 2 priority lists... maybe not a great idea
- Can we fail in the hot-path when the queue is full? Would it make sense? I feel we should crash but we don't want SDKs to fail... better eat all the memory and let the system kill us...

sighphyre (approved):
Let's do it

PR #404, opened 2026-08-05
Replace callbacks with event dispatcher in UnleashClient and its Connectors

ppicom:
## Summary

This pull request wires the EventDispatcher to the UnleashClient and its connectors. It changes all connectors from expecting an `event_callback`
(and, in some instances, a `ready_callback`) to accepting an instance
of the newly created EventDispatcher.

### `unleash_event_callback` instance variable

After reviewing https://github.com/Unleash/unleash-python-sdk/pull/243, it seems like that property was not meant to be public.
Furthermore, it is not mentioned in any docs. However, it would be a breaking
change to remove it, and rude to any users who might be using it.

I'd clean it up otherwise.

### events.py utilities

Testing a dispatcher of events that uses Threads is hard. I've introduced
an `EventRecorder` helper class that aids in testing that the publishing
of events works as intended without having to poke the internals of
the class being tested.

I agree, though, that it can be funny. I'm happy to review this decision if
better alternatives come to mind.

### asserting awaited signals

Because the callback ran in the same thread as tests, ensuring signals
were awaited instead of timed out made sense:

```
    ready_signal.wait(timeout=1) # We don't care about the return value
```

Callbacks run in another thread now, so asserting these waits makes
more sense:

```
    assert ready_signal.wait(timeout=WAIT_TIMEOUT)
```

### `BoostrapConnector` behaves funny

Two things worth noting about this connector:

1. It does not emit events because it never accepted a callback in the
first place.
2. It's the only connector which `start()` function is called in the
`UnleashClient` constructor.

### DIFF
--- a/UnleashClient/clients/unleash_client.py
+++ b/UnleashClient/clients/unleash_client.py
@@ -38,6 +38,7 @@ from UnleashClient.constants import (
 )
 from UnleashClient.events import (
     BaseEvent,
+    EventDispatcher,
     UnleashEvent,
     UnleashEventType,
     UnleashReadyEvent,
@@ -85,6 +86,11 @@ def build_ready_callback(
 ) -> Optional[Callable]:
     """
     Builds a callback function that can be used to notify when the Unleash client is ready.
+
+    .. deprecated::
+        READY is now emitted through :class:`UnleashClient.events.EventDispatcher`,
+        which deduplicates it itself.  This helper is retained for backwards
+        compatibility and is no longer used internally.
     """
 
     if not event_callback:
@@ -137,7 +143,7 @@ class UnleashClient:
     :param scheduler: Custom APScheduler object.  Use this if you want to customize jobstore or executors.  When unset, UnleashClient will create it's own scheduler.
     :param scheduler_executor: Name of APSCheduler executor to use if using a custom scheduler.
     :param multiple_instance_mode: Determines how multiple instances being instantiated is handled by the SDK, when set to InstanceAllowType.BLOCK, the client constructor will fail when more than one instance is detected, when set to InstanceAllowType.WARN, multiple instances will be allowed but log a warning, when set to InstanceAllowType.SILENTLY_ALLOW, no warning or failure will be raised when instantiating multiple instances of the client. Defaults to InstanceAllowType.WARN
-    :param event_callback: Function to call if impression events are enabled.  WARNING: Depending on your event library, this may have performance implications!
+    :param event_callback: Function to call if impression events are enabled.  Called on a dedicated background thread, so it must not rely on thread local state from the caller.
     :param experimental_mode: Optional dict to configure mode. Use {"type": "streaming"} to enable streaming or {"type": "polling"} (default).
     :param sdk_flavor: Optional identifier of an integration built on top of this SDK (e.g. an OpenFeature provider). Sent in the register + metrics payloads alongside sdkVersion so adoption of the integration can be tracked. Leave unset for plain SDK usage.
     :param sdk_flavor_version: Optional version of the integration named by sdk_flavor.
@@ -205,7 +211,11 @@ class UnleashClient:
         self.unleash_project_name = project_name
         self.unleash_verbose_log_level = verbose_log_level
         self.unleash_event_callback = event_callback
-        self._ready_callback = build_ready_callback(event_callback)
+        # Events are handed to the dispatcher, which delivers them to the user's
+        # callback on its own thread.  The callback is never called from here.
+        self.__events: Optional[EventDispatcher] = (
+            EventDispatcher(event_callback) if event_callback is not None else None
+        )
         self.connector_mode: ExperimentalMode = experimental_mode or {"type": "polling"}
         self._lifecycle_lock = threading.RLock()
         self._closed = threading.Event()
@@ -361,7 +371,7 @@ class UnleashClient:
                         url=self.unleash_url,
                         headers=base_headers,
                         request_timeout=self.unleash_request_timeout,
-                        ready_callback=self._ready_callback,
+                        events=self.__events,
                         custom_options=self.unleash_custom_options,
                     )
                 elif fetch_toggles:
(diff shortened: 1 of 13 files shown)


docs/eventcallbacks.rst
```diff
@@ -33,3 +33,24 @@ Example code using `blinker <https://github.com/pallets-eco/blinker>`_:
...
+the thread that called ``is_enabled()`` or ``get_variant()``. This means:
+
+* **Calls don't block on your callback.** ``is_enabled()`` returns as soon as the
+  event is queued, so a slow callback can't slow down flag evaluation. It also
+  means the call returns *before* your callback has run - if you need to assert on
+  an event in a test, wait for it rather than checking immediately afterwards.
+* **Thread local state from the caller isn't available.** Flask's ``g``, the
+  current Django request, ``contextvars`` and similar will not be set. Read
+  anything you need from the event itself, or capture it before the call.
+* **Callbacks are serialised.** Events are delivered one at a time, in order, so
```
sighphyre: It's probably technically correct but I see serialise and I immediately wonder why these need to be written to JSON before being sent

```suggestion
* **Callbacks run sequentially.** Events are delivered one at a time, in order, so
```

docs/eventcallbacks.rst
```diff
@@ -33,3 +33,24 @@ Example code using `blinker <https://github.com/pallets-eco/blinker>`_:
...
+  your callback doesn't need to be thread safe against itself.
+* **Events are dropped if you can't keep up.** Up to 10,000 events are held while
```
sighphyre: We changed this didn't we?

```suggestion
* **Events are dropped if you can't keep up.** Up to 100 events are held while
```

docs/eventcallbacks.rst
```diff
@@ -33,3 +33,24 @@ Example code using `blinker <https://github.com/pallets-eco/blinker>`_:
...
+  waiting on your callback; beyond that, events are discarded and a warning is
+  logged. Exceptions raised by your callback are logged and otherwise ignored.
+* **Call** ``destroy()`` **when you're done.** It delivers whatever is still
```
sighphyre, line 55: Can't remember what we did here, but my memory says we changed this behavior too

UnleashClient/clients/unleash_client.py
```diff
@@ -205,7 +211,11 @@ def __init__(  # noqa: PLR0913, PLR0917
         self.unleash_project_name = project_name
         self.unleash_verbose_log_level = verbose_log_level
         self.unleash_event_callback = event_callback
-        self._ready_callback = build_ready_callback(event_callback)
+        # Events are handed to the dispatcher, which delivers them to the user's
+        # callback on its own thread.  The callback is never called from here.
+        self.__events: Optional[EventDispatcher] = (
```
sighphyre, line 216: Would `__event_dispatcher` be a better name here?

UnleashClient/connectors/base_connector.py
```diff
@@ -32,6 +39,32 @@ def start(self):
     def stop(self):
         pass
 
+    def emit(self, event: BaseEvent) -> None:
+        """
+        Hands an event to the dispatcher, if the user asked for events at all.
+        Emitting never raises: connectors shouldn't have to guard their own
+        control flow against a failing event.
+        """
```
sighphyre, line 47: Nice!

UnleashClient/connectors/bootstrap_connector.py
```diff
@@ -12,10 +12,14 @@ def __init__(
...
-        self.engine = engine
-        self.cache = cache
         self.job = None
 
+    # TODO: this connector is never given an EventDispatcher, so bootstrapping
+    # does not emit a READY event. Bootstrapped clients only see READY once
+    # initialize_client() builds a polling, streaming or offline connector.
+    #
+    # This call to start() might need to be moved from ``UnleashClient``'s
+    # constructor to ``initialize_client``.
```
sighphyre, line 22: This is fair. I honestly don't think this entity should exist at all so right now, TODO is totally fine here

sighphyre (approved):
Yeah this is really nice

UnleashClient/clients/unleash_client.py
```diff
@@ -611,26 +622,24 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
...
+                        enabled=bool(variant["enabled"]),
+                        feature_name=feature_name,
+                        variant=str(variant["name"]),
+                    )
                 )
+        except Exception as excep:
+            LOGGER.log(
+                self.unleash_verbose_log_level,
+                "Error emitting impression event: %s",
+                excep,
```
gastonfournier, line 641: Should we create the event earlier so when we log this we can have the event logged? I think the runtime exception will lack the details of the event itself

UnleashClient/clients/unleash_client.py
```diff
@@ -561,23 +575,20 @@ def is_enabled(
...
+                        feature_name=feature_name,
+                    )
                 )
-
-                self.unleash_event_callback(event)
         except Exception as excep:
             LOGGER.log(
                 self.unleash_verbose_log_level,
-                "Error in event callback: %s",
+                "Error emitting impression event: %s",
```
gastonfournier, line 591: Same comment as below about logging the actual event

gastonfournier (commented):
Looks clean!

docs/eventcallbacks.rst
ppicom: Ah, damn. I corrected the code and forgot about these .rst files 🤦

UnleashClient/clients/unleash_client.py
ppicom: FIGHT ME!
Nah, just joking. Yeah, totally 👍

UnleashClient/connectors/base_connector.py
ppicom: Thanks!

UnleashClient/connectors/bootstrap_connector.py
ppicom: yeah... I mean, this is the only connector for which we call `start()` in the client's constructor.

I agree we can work on this in the future if it becomes a problem.

UnleashClient/clients/unleash_client.py
```diff
@@ -513,6 +522,11 @@ def destroy(self) -> None:
                 except Exception as exc:
                     LOGGER.warning("Exception during cache teardown: %s", exc)
 
+            # Closed last: the scheduler has to drain first, otherwise an
+            # in-flight poll can still queue events we'd silently drop.
+            if self.__events:
+                self.__events.close()
```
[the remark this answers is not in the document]
ppicom: yeah, you're right, robot 👍

PR #406, opened 2026-08-06
feat!: migrate to bindings v2

ppicom:
## Summary

After releasing https://github.com/Unleash/yggdrasil-bindings/pull/117, it's now the SDK's turn to use the new version of the bindings.

This new version has moved the `count_toggle`, `count_variant` and `should_emit_impression_event` inside the public methods `is_enabled` and `get_variant. As a consequence, anyone using this new version of the bindings does not need to (in fact, they must not) invoke `count_toggle` or any other helper; just `is_enabled` and `get_variant`

### DIFF
--- a/UnleashClient/clients/unleash_client.py
+++ b/UnleashClient/clients/unleash_client.py
@@ -28,7 +28,6 @@ from UnleashClient.connectors import (
 )
 from UnleashClient.constants import (
     APPLICATION_HEADERS,
-    DISABLED_VARIATION,
     ETAG,
     METRIC_LAST_SENT_TIME,
     REQUEST_RETRIES,
@@ -535,17 +534,6 @@ class UnleashClient:
         redacted_secret = f"{secret[:6]}...{secret[-3:]}"
         return f"{prefix}{separator}{redacted_secret}"
 
-    @staticmethod
-    def _get_fallback_value(
-        fallback_function: Callable, feature_name: str, context: dict
-    ) -> bool:
-        if fallback_function:
-            fallback_value = fallback_function(feature_name, context)
-        else:
-            fallback_value = False
-
-        return fallback_value
-
     # pylint: disable=broad-except
     def is_enabled(
         self,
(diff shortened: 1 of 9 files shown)


UnleashClient/impact_metrics.py
```diff
@@ -64,10 +64,10 @@ def observe_histogram(
         self._engine.observe_histogram(name, value, labels)
 
     def _variant_label(self, flag_name: str, context: Dict[str, Any]) -> str:
-        variant = self._engine.get_variant(flag_name, context)
-        if variant and variant.enabled:
+        variant = self._engine.get_variant(flag_name, context).variant
```
ppicom: With the new bindings dealing internally with `count_toggle` and `count_variant` inside of `is_enabled` and `get_variant`, now these will be counted in the metrics of the SDK as an additional variant evaluation.

Noting it here so I don't forget when we come back to this PR.

UnleashClient/clients/unleash_client.py
```diff
@@ -641,6 +620,8 @@ def get_variant(self, feature_name: str, context: Optional[dict] = None) -> dict
                 excep,
             )
 
+        # This can probably become a to_dict method of the Variant type.
+        variant = {k: v for k, v in asdict(result.variant).items() if v is not None}
```
sighphyre, line 624: Fair. I think this can live here for now. Be suuuuper cool to have a proper type bubbled here in the future and I think this work makes that easier so I think we can leave this alone until we do a major version

UnleashClient/impact_metrics.py
sighphyre: Bugger. This isn't going to work. Impact metrics can't fiddle with metrics like this

sighphyre (approved):
Okay so we talked offline. We need a good way to check the variant in impact metrics without counting. Seen the PR on the bindings 👍. Going to put an approve on this and trust you

UnleashClient/clients/unleash_client.py
ppicom: fair enough! 😄

UnleashClient/impact_metrics.py
ppicom: Solved. In the latest commit, we use check_variant instead of get_variant to find the labels of the feature.

PR #407, opened 2026-08-10
Fix/redact instance id in identifier

RaymondBibby:
## Description

Fixes three locations where `instanceId` is logged in plain text. When
`instance_id` is used as an auth token (as required by GitLab's built-in
Unleash-compatible API), these log lines expose the token in application logs.

The `Authorization` custom header (used for self-hosted Unleash auth) is
already redacted via `_redact_to_print_safely()`. This PR applies the same
protection to `instanceId` consistently across all log sites.

Changes:
- Redacts `instanceId` in `__get_identifier()` (duplicate-instance warning)
- Extracts `_redact_to_print_safely()` into `utils.py` to avoid a circular
  import from `api/register.py` and `api/metrics.py`. The class method is
  kept as a thin wrapper - no public API change.
- Redacts `instanceId` in the logged registration payload (`api/register.py`)
- Redacts `instanceId` in the logged metrics payload (`api/metrics.py`)

Wire behaviour is unchanged - the actual HTTP request bodies sent to the
server are unaffected.

Fixes # (no existing issue - discovered during integration with GitLab's Unleash API)

## Type of change

## How Has This Been Tested?

- [x] Unit tests

New tests verify that `instanceId` does not appear in plain text in logs
for all three affected sites, and that ill present
in the actual HTTP request body sent to the server.

## Checklist

- [x] I have performed a self-review of
- [x] I have added tests that prove my

### DIFF
--- a/UnleashClient/clients/unleash_client.py
+++ b/UnleashClient/clients/unleash_client.py
@@ -51,6 +51,7 @@ from UnleashClient.utils import (
     InstanceAllowType,
     InstanceCounter,
     extract_environment_from_headers,
+    redact_to_print_safely,
 )
 
 try:
@@ -515,11 +516,7 @@ class UnleashClient:
 
     @staticmethod
     def _redact_to_print_safely(value: Optional[str]) -> Optional[str]:
-        if not value:
-            return value
-        prefix, separator, secret = value.rpartition(":")
-        redacted_secret = f"{secret[:6]}...{secret[-3:]}"
-        return f"{prefix}{separator}{redacted_secret}"
+        return redact_to_print_safely(value)
 
     @staticmethod
     def _get_fallback_value(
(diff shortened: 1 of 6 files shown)


UnleashClient/clients/unleash_client.py
```diff
@@ -515,11 +516,7 @@ def destroy(self) -> None:
 
     @staticmethod
     def _redact_to_print_safely(value: Optional[str]) -> Optional[str]:
```
sighphyre, line 518: I think maybe we should leave this alone and keep it responsible just for redacting the Unleash tokens. I say this because this is completely pinned to redacting the Unleash token formats and those may change in the near future, which would mean we risk breaking Gitlab's constraints again

UnleashClient/utils.py
```diff
@@ -63,6 +63,14 @@ def get_identifier(context_key_name: str, context: dict) -> Any:
     return value
 
 
+def redact_to_print_safely(value: Optional[str]) -> Optional[str]:
```
sighphyre, line 66: Could we maybe make a new function for this that operates just on instance strings rather than also trying to do the API tokens too? `_redact_header` or something? And while we're here, any chance of underscoring it as well - I know historically we've been bad at doing that in this SDK but we're trying to be better

PR #409, opened 2026-08-18
EG-4662 (1): Scaffold of new `AsyncUnleashClient` + extraction of `UnleashConfig`

ppicom:
## Quick summary

---

> [!NOTE]
> This being the first PR of the stack, the summary won't be as quick as in the ones to follow. Here I lay out the idea behind the stack of PRs to come.

The previous stacked focused on threaded work to pull event callbacks out of the hot path. This on is directly concerned with the implementation of AsyncUnleashClient by reusing as much code as possible.

`UnleashClient` was doing too many things. Moving them to an `ABC` that both clients shared would make things worse in the future and, because one had to be sync and the other did not, it made sharing code very hard. So, instead of inheritance I've leaned on _composition_: I've broken down each responsibility into a separate class and then used that new class in both flavors of client (`UnleashClient` and `AsyncUnleashClient`).

I'm still working on the classes that cannot be shared between sync and async implementations (aka the colored classes).

At the moment, `UnleashClient` is a single class holding fifteen responsibilities:

1. Constructor argument normalization and validation.
2. Client identity: connection id, app name, instance id, header assembly.
3. Duplicate-instance detection against a module-global counter.
4. Ownership of the Yggdrasil engine and flag evaluation.
5. Context enrichment and sanitization.
6. Impression event construction.
7. Cache ownership and seeding.
8. Bootstrap from cache.
9. Registration with the Unleash server.
10. Feature provisioning strategy selection.
11. Job scheduling via APScheduler.
12. The metrics send loop and the shutdown flush.
13. Impact metrics wiring.
14. Run-state machine, locking, and shutdown ordering.
15. HTTP concerns: `custom_options`, timeouts, retries.

Our goal is to extract them one by one (more or less, some like shutdown ordering do make sense in the parent class) into separate objects that have a single responsibility. Once a responsibility is out of `UnleashClient` and into another object (e.g. `UnleashConfig`), that new object can be shared between the two flavors of the client.

---

### This PR

In this pull request, I have extracted all the values of `UnleashClient` that configure the behavior of the SDK and moved them to a new `UnleashConfig` class.

That class is then a collaborator that both the `UnleashClient` and the `AsyncUnleashClient` can share.

I have then scaffolded the `AsyncUnleashClient` class, which does nothing at this point.

## Related context

## What to focus on

This PR is rather simple, I'd pay attention to not having forgotten any configuration value in the process.

## How to Verify / Test

- `make test` should pass

## Risks & Operational Notes

- [ ] We have not lost any configuration value that users could use to configure their SDK.

### DIFF
--- a/UnleashClient/clients/unleash_client.py
+++ b/UnleashClient/clients/unleash_client.py
@@ -19,6 +19,11 @@ from yggdrasil_engine.engine import UnleashEngine
 
 from UnleashClient.api import register_client
 from UnleashClient.cache import BaseCache, FileCache
+from UnleashClient.config import (
+    ExperimentalMode,
+    UnleashConfig,
+    redact_to_print_safely,
+)
 from UnleashClient.connectors import (
     BaseConnector,
     BootstrapConnector,
@@ -50,13 +55,7 @@ from UnleashClient.utils import (
     LOGGER,
     InstanceAllowType,
     InstanceCounter,
-    extract_environment_from_headers,
 )
-
-try:
-    from typing import Literal, TypedDict
-except ImportError:
-    from typing_extensions import Literal, TypedDict  # type: ignore
 
 INSTANCES = InstanceCounter()
 _BASE_CONTEXT_FIELDS = [
@@ -74,10 +73,6 @@ class _RunState(IntEnum):
     UNINITIALIZED = 0
     INITIALIZED = 1
     SHUTDOWN = 2
-
-
-class ExperimentalMode(TypedDict, total=False):
-    type: Literal["streaming", "polling"]
 
 
 def build_ready_callback(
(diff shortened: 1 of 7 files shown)


UnleashClient/config.py
```diff
@@ -0,0 +1,98 @@
...
+def redact_to_print_safely(value: Optional[str]) -> Optional[str]:
+    if not value:
+        return value
+    prefix, separator, secret = value.rpartition(":")
+    redacted_secret = f"{secret[:6]}...{secret[-3:]}"
+    return f"{prefix}{separator}{redacted_secret}"
+
+
+@dataclass
+class UnleashConfig:
```
sighphyre, line 29: AWWW YISSS. I am here for this, think this is a nice move extracting this out into a dumb config object

UnleashClient/config.py
```diff
@@ -0,0 +1,98 @@
...
+    connection_id: str = field(default_factory=lambda: str(uuid.uuid4()))
+    static_context: Dict[str, Any] = field(init=False, default_factory=dict)
+
+    def __post_init__(self) -> None:
+        self.url = self.url.rstrip("/")
+        self.refresh_jitter = (
+            int(self.refresh_jitter) if self.refresh_jitter is not None else None
+        )
+        self.metrics_jitter = (
+            int(self.metrics_jitter) if self.metrics_jitter is not None else None
```
sighphyre, line 67: I realise you didn't create this problem, you're just moving it over. And I must have seen this code a million times but now that it's in a PR I think it's silly

If you want to leave it the hell alone I can absolutely live with that, but if you want to do something a little saner, I am totally open to not doing this - we ingest an Optional[int] at the boundary, I'm okay to trust it if you are

Think this applies to refresh_jitter as well

UnleashClient/config.py
```diff
@@ -0,0 +1,98 @@
...
+    request_retries: int = REQUEST_RETRIES
+    project_name: Optional[str] = None
+    verbose_log_level: int = 30
+    sdk_flavor: Optional[str] = None
+    sdk_flavor_version: Optional[str] = None
+    experimental_mode: Optional[ExperimentalMode] = None
+    connection_id: str = field(default_factory=lambda: str(uuid.uuid4()))
+    static_context: Dict[str, Any] = field(init=False, default_factory=dict)
+
+    def __post_init__(self) -> None:
```
sighphyre, line 61: I didn't even know this dunder method existed! Gosh I have been out of the Python game a long time. Learned a thing, cheers!

UnleashClient/config.py
```diff
@@ -0,0 +1,98 @@
...
+        self.refresh_jitter = (
+            int(self.refresh_jitter) if self.refresh_jitter is not None else None
+        )
+        self.metrics_jitter = (
+            int(self.metrics_jitter) if self.metrics_jitter is not None else None
+        )
+        # `or` rather than `is None`, so that passing {} yields a fresh dict,
+        # as the client constructor has always done.
+        self.custom_headers = self.custom_headers or {}
+        self.custom_options = self.custom_options or {}
```
sighphyre, line 72: Think I'm having a dumb here - can't we just use 

``` python
custom_headers: dict = field(default_factory=dict)
custom_options: dict = field(default_factory=dict)
```

In the constructor?

UnleashClient/clients/unleash_client.py
```diff
@@ -290,13 +272,173 @@ def _init_scheduler(
...
+    def unleash_sdk_flavor_version(self, value: Optional[str]) -> None:
+        self._config.sdk_flavor_version = value
+
+    @property
+    def connector_mode(self) -> ExperimentalMode:
+        return self._config.experimental_mode
+
+    @connector_mode.setter
+    def connector_mode(self, value: ExperimentalMode) -> None:
+        self._config.experimental_mode = value
```
sighphyre, line 433: Whoo boy, that's a lot of setters and getters!

How I feel about this depends on your/our plans here.

- If this is a shim to get us past this PR and we'll eventually start moving access to this data to the config object then yeah I think this is fine
- If this is a longer term solution where we're leaving this in place because they're kinda sorta public fields and we want to deprecate access here then I think it's okay
- If we're married to this forever then I hate it and I think we should some kind of proxy decorator class here because maintaining this is gonna suck. Don't want to discuss that option unless we're definitely committed to leaving this code here forever

To be clear - we can also just merge this and do a decorator here later, which I'm more in favor of than fiddling right now

sighphyre (approved):
Cheers. I think this is solid. Love the new Config object. I have some thoughts, most of it is related to icky legacy code that might want to be taken out back and shot

But to be clear - I think this is good and I'm happy merging as is without fiddling

UnleashClient/config.py
ppicom: If you're here for dumb code, you're at the right place! haha

thanks 🤟

UnleashClient/config.py
ppicom: yeah, forcing a client to now input an int felt like a breaking change. Kind of like changing the type of a field in an HTTP request field. Soooo, I worked around it with this.

But, unless we change the value of the argument of the client, this code still needs to exist I think 🤔 

moving it to the boundary means having it twice, though 🙌 

I'mma note it as something that we can figure out in a subsequent release 👍 it makes a lot of sense

UnleashClient/config.py
ppicom: I honestly did not know, but research shows that:

> `field()` outside a class body just returns a `dataclasses.Field` instance, so `self.custom_headers = field(default_factory=dict)` assigns a `Field` object, not a `dict`. It fails silently until something tries to use it as a mapping.

ppicom: It's a "if the constructor did not set this, set them to empty dicts" sanity check in the post init step, if that makes it clearer.

A different phrasing of the same:

```python

if not self.custom_headers:
    seld.custom_headers = {}
```

PR #410, opened 2026-08-18
EG-4662 (2): Extract `ContextEnricher`

ppicom:
## Quick summary

This PR moves the building of the context to a `ContextEnricher` class and turns it into a collaborator that both the `UnleashClient` and its async homonymous share.

## Related Context

## Approach

Until now, `UnleashClient` would deal with grabbing the context from the user (a plain `dict` with arbitrary key value pairs) and reshaping it into the context that the unleash servers expect: a dictionary with the keys: `"userId", "sessionId", "environment", "appName", "currentTime", "remoteAddress", "properties"` at its first level, and the context from the user relegated to the `properties` map.

This logic has been moved to its own class and it is now shared between the two clients.

## Where to focus the review

- `UnleashClient/context.py` must return the context in the same shape that existed before.

## How to Verify

- `make test` should pass

## Risks & Operational Notes

- [ ] This PR moves the logic that builds the context. Breaking it renders the SDK unusable.

### DIFF
--- a/UnleashClient/clients/async_unleash_client.py
+++ b/UnleashClient/clients/async_unleash_client.py
@@ -1,9 +1,10 @@
 """
 Asynchronous Unleash client.
 
-Work in progress.  This class currently only builds the shared
-:class:`UnleashClient.config.UnleashConfig`; it performs no I/O and is not
-exported from the package root.  See ``docs/object-composition.md``.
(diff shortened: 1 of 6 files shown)


sighphyre (approved):
No notes, love it

PR #413, opened 2026-08-18
EG-4662 (4): Extract application state keeping to `FeatureStore` + drill `unleash_refresh_jitter` to `PollingConnector`

ppicom:
## Quick Summary

This PR moves state-keeping (different from state-fetching) out of `UnleashClient` and its `*Connector` collaborators, and into a separate `FeatureStore` class. Then, shares that class with `AsyncUnleashClient` as well.

Also, when sharing the new collaborator, it was straightforward to fix a bug: `unleash_refresh_jitter` was never passed down to the `PollingConnector`. Therefore, the value from the user had no effect.

## Related Context

## Approach

- **Problem:** Among other things (see #409), the `UnleashClient` dealt with keeping the state of the application. Not only that, the Connectors also have some logic regarding what is the state of the application. This makes it hard to test and impossible to share across two client implementations.
- **Solution:** Move the logic that keeps track of the SDK state into a single class: `FeatureStore`. Note one important thing: this class is concerned with "applying" state, not fetching it. In other words, owns what happens to feature state once it has arrived. That is how we've managed to keep it un-colored (i.e. not async).
  There is one method per source rather than a single ``apply``, because the three steps happen in a different order, over different payloads, with different failure handling depending on where the state came from.

## Reviewer Guide

- `UnleashClient/store.py`: Holds the implementation of the new store. `apply_fetched` is concerned with storing state that came from a fetching connector, `apply_streamed` is concerned with storing state that was streamed.
- `UnleashClient/clients/unleash_client.py`: Changes to the current implementation of the client to rely on the new `FeatureStore`.
- `UnleashClient/connectors/*`: Connectors also change to accommodate for the new FeatureStore.

## How to Verify / Test

You probably guessed it:

```sh
make test
```

## Risks & Operational Notes

- [ ] Changes to the logic on how we hand the state over to the engine and cache it on file is not meant to have changed - just move.
- [ ] `appy_fetched()` and `apply_streamed()` should be enough public API for the FeatureStore. `load_from_cache` and `emit_ready` leak part of the responsibility to collaborators.
- [ ] `UnleashClient` had two properties: `self.cache` and `self.engine`. They are not documented as public, and checking the PR that introduced them, it seemed like they were meant to be private. I've made them private (i.e. prepended an `_` underscore) to make it explicit. BUT that COULD technically be a breaking change.


UnleashClient/store.py
```diff
@@ -0,0 +1,158 @@
...
+    BaseEvent,
+    EventDispatcher,
+    UnleashEventType,
+    UnleashFetchedEvent,
+    UnleashReadyEvent,
+)
+from UnleashClient.utils import LOGGER
+
+
+class FeatureStore:
```
sighphyre, line 20: This is very entertaining. This is an abstraction I've quietly started to slip into SDKs when I'm already touching them in aggressive ways. It's just kinda vindicating to see someone do exactly the same thing without being told 

I think what this is closing over is not exactly what I would like right now - not a disaster, it can't  be injected so we can consider it private and iterate on it later - and this isn't something you could have foreseen because the Python SDK is doing something weird. But... right now we're communicating with bootstrap and filebackup via the cache. That's an annoying implementation detail of fcache, and something I want to get away from in the future. That, in turn, is only done because bootstrap is tied to fcache and it shouldn't be. Ideally this would close over an abstract backup source (with only one implementation - a file backup) and a bootstrapper that can be user defined.

Don't need to change anything, just sharing some oddities about this particular SDK so we can fix it together in the future (and by we, yes I mean you and me)

UnleashClient/store.py
```diff
@@ -0,0 +1,158 @@
...
+
+        Warns and returns when the cache is empty, so nothing is emitted and the
+        engine keeps whatever state it already had.  Engine failures are logged
+        with the offending body at debug level, never raised: every caller is
+        already falling back on the cache and has nothing better to do with the
+        failure.
+        """
+        feature_provisioning = self._cache.get(FEATURES_URL)
+        if not feature_provisioning:
+            LOGGER.warning(
```
sighphyre, line 62: Think this should be an info level log. Loading from cache should be polite but not necessary

sighphyre (approved):
Nice work!

UnleashClient/store.py
ppicom: Thanks, I appreciate the context a lot.

> Don't need to change anything, just sharing some oddities about this particular SDK so we can fix it together in the future (and by we, yes I mean you and me)

Awesome! I'm down for that. I've merged and archived this issue so we don't lose track of it 👍

UnleashClient/store.py
ppicom: Oh, you're right. I've already merged this one (shame on me). But I will sneak this change in the PRs to come 👍

PR #414, opened 2026-08-18
EG-4662 (5): Extract `Scheduler`

ppicom:
## Quick Summary

This PR extracts all logic related to scheduling jobs and all references to `apscheduler` into a single collaborator: `Scheduler`.

Then, it refactors `UnleashClient` to use it and it wires it up in `AsyncUnleashClient`.

## Related Context

## Approach (Why & How)

- **Problem:** Scheduling logic was spread out, many different modules were depending directly on `apscheduler`. This made it harder to share and to test.
- **Solution:** One option was to write code in `AsyncUnleashClient` that used `apscheduler` directly. However, extracting it to a collaborator, reducing the surface of dependency with the library and writing the code just once seemed to be the better alternative for maintainability and feature parity in the future.

## Reviewer Guide (Where to Focus)

- `UnleashClient/scheduler.py`: implementation of `Scheduler` using `ThreadPoolExecutor` and `BackgroundScheduler`
- `UnleashClient/connectors/offline_connector.py` and `UnleashClient/connectors/polling_connector.py`: no longer use `apscheduler` directly.

## 🧪 How to Verify / Test

```sh
make test
```

## ⚠️ Risks & Operational Notes
- `UnleashClient` allows the configuration of a custom scheduler and a custom executor name. This should still be possible.
- Public properties have been kept public and change-able via getters and setters.


sighphyre (approved):
Yep, it's not perfect but it solves the duplication problem. Let's do it!

PR #415, opened 2026-08-18
EG-4662 (6): Extract `Transport` and wire it in the sync client

ppicom:
## Quick Summary

This PR moves the HTTP work to a separate collaborator: the `Transport`. It supersedes the preceding `/api` package, and uses the `HeaderFactory` and the `UnleashConfig` to abstract away the work of bringing the state of features from the remote server.

Note that this collaborator is colored - it won't be shared between the two clients. But it helps at keeping "branched" code at a minimum.

## Related Context

Contributes to #240 

## Approach

- **Problem:** HTTP work was all inside of the `/api` package, making it hard to abstract away and/or reuse. Now we need an async version of what `/api` was doing.
- **Solution:** Instead of copying and pasting `/api`, slapping an `async` keyword to the function definitions and being done with it, I've reworked the HTTP work so that all the "sync" work is shared (payload and header building) and only the actual request differs. For that, the `UnleashClient` now uses a `Transport` (spoiler: the `AsyncUnleashClient` will use a `AsyncTransport`). This is one of the leaves 🍃  that is colored, and hence cannot be shared between flavors.

## 🛠️ Reviewer Guide (Where to Focus)

- `UnleashClient/transport.py`: the new implementation of what `/api` was doing.

## 🧪 How to Verify / Test

```sh
make test
```

## ⚠️ Risks & Operational Notes

N/A


UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+                    **config.custom_options,
+                )
+
+            if resp.status_code not in [200, 304]:
+                _log_resp_info(resp)
+                LOGGER.warning(
+                    "Unleash Client feature fetch failed due to unexpected HTTP status code: %s",
+                    resp.status_code,
+                )
+                raise Exception(
```
ppicom, line 110: Probably worth using a typed exception. `TransportException` for example.

UnleashClient/payloads.py
```diff
@@ -0,0 +1,46 @@
...
+from platform import python_implementation, python_version
+from typing import Any, Dict
+
+import yggdrasil_engine
+
+from UnleashClient.config import UnleashConfig
+from UnleashClient.constants import CLIENT_SPEC_VERSION, SDK_NAME, SDK_VERSION
+
+
+def build_register_payload(
```
ppicom, line 13: I'm compelled to also turn this into a class, like headers.py.

UnleashClient/transport.py
sighphyre: It's only useful if you're planning on doing something with it. Don't think that's the case here, any exception here is going to be bounded by the try/catch and will result in the same logging block. I think it's fine

UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+        config = self._config
+        try:
+            LOGGER.info("Getting feature flag.")
+
+            # The features endpoint has always been sent an uppercase copy of
+            # the two identification headers, on top of the lowercase pair in
+            # HeaderFactory.base().  It stays here rather than in polling()
+            # because it is a quirk of this one endpoint, and because
+            # test_polling_and_metrics_differ_only_in_the_interval pins
+            # polling() against exactly this kind of addition.
```
sighphyre: This sounds like Claude thinking out loud. Is it useful for future maintainers? I'm not sure it is

UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+
+            adapter = HTTPAdapter(
+                max_retries=Retry(
+                    total=config.request_retries, status_forcelist=[500, 502, 504]
+                )
+            )
+            # A session per fetch, as the api/ function did. One Transport is
+            # reachable from the polling thread, the metrics thread and
+            # destroy()'s caller at once, and requests.Session is not documented
+            # as thread-safe.
```
sighphyre: Think this is more Claude musing, not actually sure what it's saying here or what this would imply for someone mucking around in this file

UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+                    "Unleash Client registration failed due to unexpected HTTP status code: %s",
+                    resp.status_code,
+                )
+                return False
+
+            LOGGER.info("Unleash Client successfully registered!")
+
+            return True
+        # Ahead of the RequestException clause below: all four subclass it, and
+        # Python matches the first clause that fits.
```
sighphyre: Claaaaaaaaauuuude, go make a mess somewhere else

UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+    that an async twin duplicates the sending and nothing else.
+
+    Each method asks the :class:`~UnleashClient.headers.HeaderFactory` for the
+    header set its endpoint needs; callers pass none. Both the factory and the
+    config are read on every request rather than captured, because
+    ``unleash_url``, ``unleash_custom_options``, ``unleash_request_timeout``,
+    ``unleash_request_retries``, ``unleash_project_name``,
+    ``unleash_custom_headers``, ``unleash_app_name`` and ``unleash_instance_id``
+    are all public and writable.
+
```
sighphyre: I think we should take this addendum out. It's useful context for this PR, but once this is merged, the context that makes it useful is gone

UnleashClient/transport.py
```diff
@@ -0,0 +1,249 @@
...
+    ``raw_state`` is None both when the server answered 304 and when the fetch
+    failed outright; ``not_modified`` is what tells the two apart. Nothing acts
+    on that distinction yet — :class:`UnleashClient.store.FeatureStore` reloads
+    from the cache in either case, as it always has.
+    """
+
+    raw_state: Optional[str]
+    etag: str
+    not_modified: bool = False
```
sighphyre, line 27: Why do we need both not_modified and a raw_state? Previously we returned a 2 tuple of raw_state and etag. Does this solve a problem I'm not seeing?

UnleashClient/payloads.py
```diff
@@ -0,0 +1,46 @@
...
+        "sdkVersion": f"{SDK_NAME}:{SDK_VERSION}",
+        "strategies": [*strategies],
+        "started": datetime.now(timezone.utc).isoformat(),
+        "interval": config.metrics_interval,
+        "platformName": python_implementation(),
+        "platformVersion": python_version(),
+        "yggdrasilVersion": yggdrasil_engine.__yggdrasil_core_version__,
+        "specVersion": CLIENT_SPEC_VERSION,
+    }
+    # Only sent when the client was configured with an integration flavor
```
sighphyre: Yup, that's how if statements work. Don't think this is necessary

```suggestion

```

UnleashClient/payloads.py
sighphyre: Can if it solves a problem. But classes cause new problems so unless it does solve something I think a function is fine. Module level functions are a totally respectable way of composing a Python program, don't give in to the JVM whispering in your ear

UnleashClient/payloads.py
```diff
@@ -0,0 +1,46 @@
...
+
+def build_register_payload(
+    config: UnleashConfig, strategies: Dict[str, Any]
+) -> Dict[str, Any]:
+    """
+    Build the body of a client registration request.
+
+    Called immediately before the request, as it was when this dict lived at the
+    top of ``register_client``: ``started`` is the moment of the call, not the
+    moment the client was constructed.
```
sighphyre: Yep, technically true without communicating intent

```suggestion
    top of ``register_client``: ``started`` is the time of registration.
```

UnleashClient/connectors/polling_connector.py
```diff
@@ -12,51 +12,28 @@ def __init__(
...
-        request_timeout: int = 30,
-        request_retries: int = 3,
-        project: Optional[str] = None,
+        transport: Transport,
         refresh_interval: int = 15,
         refresh_jitter: Optional[int] = None,
     ):
         """
+        :param transport: Performs the fetch. Owns the url, timeout, retries,
+                          project, custom options and headers it used to be handed.
```
sighphyre: Maybe? Helps understand what it's supposed to do, not just what it does

```suggestion
        :param transport: Performs the fetch. Owns everything needed to perform  
                          HTTP communication.
```

UnleashClient/clients/unleash_client.py
```diff
@@ -518,16 +504,15 @@ def initialize_client(self, fetch_toggles: bool = True) -> None:
                     # be unconditional.
                     start_scheduler = True
 
+                    # Informational only since the Transport builds its own
+                    # headers; kept because it is a public attribute.
                     self.metrics_headers = self._headers.metrics()
```
sighphyre: Meh. We're removing a few things that weren't really part of the documented public API. I think this should go the same way

sighphyre (approved):
Yeah I think the change is solid. Might giving the LLM goop a once over before merging? Sorry, I've had a day with Claude today 😢

UnleashClient/payloads.py
ppicom: > don't give in to the JVM whispering in your ear

🤣 roger that!

UnleashClient/transport.py
ppicom: oh jesus, some slop in the docstrings escaped my watch 🤦

UnleashClient/transport.py
ppicom: dang it

UnleashClient/transport.py
ppicom: sheeesh, you're right, those comments stink. sorry about that 🤦

ppicom: goop: gone ✅

UnleashClient/transport.py
ppicom: goop: gone ✅

UnleashClient/payloads.py
ppicom: yep, removed that trash, too

PR #420, opened 2026-08-19
EG-4662 (9): Extract flag evaluation to common `Evaluator`

ppicom:
- **chore: install aiohttp as an optional-dependency**
- **feat: Implement new `AsyncTransport`**
- **feat: wire AsyncUnleashClient with an AsyncTransport**
- **feat: extract feature evaluation into a common Evaluator**
- **refactor: delegate feature evaluation to Evaluator in UnleashClient**
- **feat: wire AsyncUnleashClient with an Evaluator**
- **docs: tentative changelog update**


tests/unit_tests/clients/test_async_unleash_client.py
```diff
@@ -254,3 +254,118 @@ def test_constructing_the_async_client_does_not_start_the_scheduler(tmpdir):
...
+
+
+def test_async_client_builds_a_transport_over_its_config_and_headers(tmpdir):
+    client = build_async_client(tmpdir, url=URL, app_name=APP_NAME)
+
+    assert client._transport._config is client._config
+    assert client._transport._headers is client._headers
+
+
+def test_constructing_the_async_client_opens_no_session(tmpdir):
```
ppicom: Applying some of the feedback I got on previous PRs, there are tests here that are to bound to the internals of the class and not really bringing any value.

I'll review the test suite and delete some of them.

UnleashClient/clients/async_unleash_client.py
```diff
@@ -95,9 +97,66 @@ def __init__(  # noqa: PLR0913, PLR0917
...
+        feature_name: str,
+        context: Optional[dict] = None,
+        fallback_function: Callable = None,
+    ) -> bool:
+        """
+        Checks if a feature toggle is enabled.
+
+        Notes:
+
+        * If client hasn't been initialized yet or an error occurs, flag will default to false.
```
sighphyre: Default is the result of the fallback function rather than False, surely?

Not important - this can change, just checking I'm understanding the plan

UnleashClient/evaluator.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class VariantResult(NamedTuple):
+    """The variant a lookup resolved to, and whether the engine knew the toggle."""
+
+    variant: Dict[str, Any]
+    is_found: bool
+
+
+class Evaluator:
```
sighphyre: Not sure how I feel about this abstraction to be honest. It takes us forward though so I want to merge and move on. However, if we make this `_Evaluator`  rather than `Evaluator` we can come back and muck with it, guilt free, when we're happier that we understand what the important abstractions are. Benefit of hindsight and all that

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
 * (Minor): `AsyncUnleashClient` ignores `custom_options`. The constructor still accepts it and it still reaches the shared configuration, but the async request path never passes it to the HTTP library, so the async client's surface does not depend on aiohttp's own keyword arguments. `UnleashClient` is unaffected and keeps passing `custom_options` to `requests`.
+* (Minor): Flag evaluation now happens through one internal `Evaluator` object instead of being implemented on the client. `is_enabled()`, `get_variant()` and `feature_definitions()` keep their signatures, their return shapes and the impression events they emit. Nothing changes for code using `UnleashClient`. The in-progress asynchronous client is wired with the same object but does not expose it yet: its `is_enabled()`, `get_variant()` and `feature_definitions()` raise `NotImplementedError` alongside the rest of its surface until it can initialize.
```
sighphyre: Is this really a minor patch? If that's really internal and doesn't change the public API, surely this is a patch?

UnleashClient/clients/unleash_client.py
```diff
@@ -214,15 +212,17 @@ def __init__(  # noqa: PLR0913, PLR0917
...
+            config=self._config,
+            events=self.__events,
+        )
+
         self._transport = Transport(self._config, self._headers)
 
         self._scheduler = Scheduler(scheduler, scheduler_executor)
 
-        if custom_strategies:
-            self._engine.register_custom_strategies(custom_strategies)
```
sighphyre, line 222: Is this an important change? We're now delaying error detection until initialize, which feels a little unfortunate because we're pushing errors right. That means users can now create a Client in a way that will never work and only know about that when they attempt to initialize it.

Not a blocker, we can change this in a patch version, just food for thought

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+* (Minor): Custom strategies are now registered on the engine by `initialize_client()` rather than by the constructor. The `ValueError` raised for a strategy that has no `apply` method, or whose `apply` does not take exactly two parameters, therefore surfaces from `initialize_client()` instead of from `UnleashClient(...)`. Code that passes valid strategies and calls `initialize_client()` before evaluating anything is unaffected; a bootstrapped client that evaluates a custom-strategy flag before initializing now gets `False` for that strategy.
```
sighphyre, line 6: Honestly I'd just call this a patch. It's probably technically a minor but I can't see it mattering ever from a public API perspective

tests/unit_tests/test_evaluator.py
```diff
@@ -0,0 +1,229 @@
+import json
```
sighphyre, line 1: Also not too important but I would suggest taking a critical eye to the value of these tests. A lot of them are just exercising the engine's responsibility indirectly

Tests can be deleted so don't let it stop you merging this

sighphyre (approved):
Yep seems sane. I would probably just take note of some of my comments, merge this and we can plan some clean up later. Right now we're still fishing for good abstractions here so any comments I make around them feeling awkward can't really be actioned until the Async Client is complete and released - just want to make sure we leave ourselves some two way doors

tests/unit_tests/test_evaluator.py
ppicom: The `Evaluator` is the dance between the engine, the enricher (unfortunate name if we go with actual context enrichers) and the events.

So, to me, they feel like social tests.

But I'm sure there's more to what you said, so I'll give it a spin in my brain. Since we can delete them without any impact, I won't block the PR because of this 👍

UnleashClient/clients/async_unleash_client.py
ppicom: yes, exactly, sorry about that. these comments should not have made it into the PR. I'm at a point where Claude is writing good code, but super shitty comments.

This should not have made it into the PR. My bad. Fixing it!

UnleashClient/evaluator.py
ppicom: I am operating under the assumption that anything in a subfolder is considered "private", and have passed on the underscore-means-private convention. If that is not the case, it makes a lotta sense to start prepending a big bunch of underscores to many many things 👍 

thanks for the thorough review 🤟

CHANGELOG.md
ppicom: I'm including everything on the CHANGELOG with the idea of refining it after we release the new version. So I would not pay excessive mind to it right now (unless there's a reason to do so that I'm not seeing 🤔 )

I take note to be more explicit about this in the future.

But yeah, I was locked on "since these aren't bug fixes, then I should bump the minor". But I like your way of seeing it better.

CHANGELOG.md
ppicom: Agreed

UnleashClient/clients/unleash_client.py
ppicom: I'd say it is - if a class comes with lifecycle methods attached to it I kinda expect the constructor to do nothing beside allocating memory and instancing classes, and then the `init()` (or similar) method to do all the preflight checks.

A user is probably initializing the client in the startup phase of its application. Buuuut it's true that this is observable behavior and it can be called a breaking change.

I just thought it would be nice to get feature parity and to get as many checks as possible out of the constructor and into `initialize_client()`

PR #421, opened 2026-08-19
EG–4662 (10): Extract metrics reporting into `MetricsReporter`

ppicom:
- **feat: abstract metrics logic away behind a new MetricsReporter class**
- **refactor: delefagate metrics collection and sending to MetricsReporter**
- **refactor: cleanup unused periodic and periodic_tasks modules**
- **docs: tentative CHANGELOG update**


UnleashClient/payloads.py
```diff
@@ -42,3 +42,37 @@ def build_register_payload(
         payload["sdkFlavorVersion"] = config.sdk_flavor_version
 
     return payload
+
+
+def build_metrics_payload(
```
alexanto, line 51: kind of minor but this fn shares code (fields and if statements) with `build_register_payload`, maybe it makes sense to extract those into another fn, wdyt?

alexanto (approved):
looks like a solid refactor to me, but I am here mainly to learn :)
minor comment, approving to unblock

UnleashClient/payloads.py
ppicom: oh, nice catch! On it 😄

PR #422, opened 2026-08-19
EG-4662 (11): Write async metrics exportation in `AsyncMetricsReporter`

ppicom:
basically, the python SDK had no support for the async keyword (big oversimplification here, but yeah).

So, we're writing an `AsyncUnleashClient` . But, instead of copy and paste, what I'm doing is extracting common behavior into smaller classes that then both clients can share

Then, it so happens that some collaborators need to be different - mostly the network related ones: in this case, the reporting of metrics requires a network call; the sync one does it through `requests` (a sync http library) and this async one does it with another one.

Note: the HTTP is abstracted away behind the "transport" abstraction.

- **feat: new AsyncMetricsReporter**
- **feat: wire AsyncMetricsReporter into AsyncUnleashClient**
- **docs: tentative update to changelog**


tests/unit_tests/test_async_metrics_reporter.py
```diff
@@ -0,0 +1,524 @@
...
+    # Computing each fire from the clock instead would push every subsequent send out by
+    # the duration of the last one, and the drift would compound.
+    fires = []
+    reporter = build_reporter(metrics_interval=30)
+
+    async def wait_until(fire):
+        fires.append(fire)
+
+    reporter._wait_until = wait_until
+    reporter.flush = _slow_flush
```
alexanto: what if the `_slow_flush` is really slow? slower than the `metrics_interval`
an empty POST is sent unecessary i think?

wdyt, would it make sense to skip the scheduled jobs in that case? maybe it is too much of an edge case

UnleashClient/async_metrics_reporter.py
```diff
@@ -0,0 +1,145 @@
...
+        bucket = self._engine.get_metrics()
+        impact_metrics = self._impact_metrics.collect()
+
+        if not (bucket or impact_metrics):
+            LOGGER.debug("No feature flags with metrics, skipping metrics submission.")
+            return
+
+        payload = build_metrics_payload(self._config, bucket, impact_metrics)
+        if not await self._transport.send_metrics(payload) and impact_metrics:
+            self._impact_metrics.restore(impact_metrics)
```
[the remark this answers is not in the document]
alexanto: if everything but 202 is a failure, if e.g the server is slow to reply we restore send the metrics again, that can mean, the server already got them and they are received duplicated.

i just wonder if this is deliberate and if sending metrics duplicated or losing them in flight is a bigger problem here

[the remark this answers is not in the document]
alexanto: could we not restore them also when there is `CancelledError` ?

UnleashClient/async_metrics_reporter.py
```diff
@@ -0,0 +1,145 @@
...
+        task, self._task = self._task, None
+        task.cancel()
+        try:
+            await task
+        except asyncio.CancelledError:
+            pass
+        except Exception as exc:
+            LOGGER.warning("Metrics reporting had already stopped: %s", exc)
+
+        await self.flush()
```
alexanto: flush is in a try catch in `_run`, but not here, why?

UnleashClient/clients/async_unleash_client.py
```diff
@@ -100,6 +107,12 @@ def __init__(  # noqa: PLR0913, PLR0917
         )
         self._transport: AsyncTransport = AsyncTransport(self._config, self._headers)
         self._scheduler: Scheduler = Scheduler()
+        self._metrics: AsyncMetricsReporter = AsyncMetricsReporter(
+            config=self._config,
+            transport=self._transport,
+            engine=self._engine,
+            impact_metrics=self.impact_metrics,
+        )
```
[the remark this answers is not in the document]
ppicom: this is on purpose -- the wiring of the client will happen once all the collaborators are in place.

UnleashClient/async_metrics_reporter.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        self._config: UnleashConfig = config
+        self._transport: AsyncTransport = transport
+        self._engine: UnleashEngine = engine
+        self._impact_metrics: ImpactMetrics = impact_metrics
+        self._task: Optional["asyncio.Task"] = None
+
+    async def start(self) -> None:
+        """Starts the recurring send on the current event loop."""
+        self._task = asyncio.create_task(self._run())
```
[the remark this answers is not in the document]
ppicom: In https://github.com/Unleash/unleash-python-sdk/pull/428 I've pulled all scheduling logic into its own class. This should be solved.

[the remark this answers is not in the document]
ppicom: Sorry about that 🦖

tests/unit_tests/test_async_metrics_reporter.py
ppicom: There's a bit of an overlap between this PR and the one in https://github.com/Unleash/unleash-python-sdk/pull/428

Mostly because I extracted scheduling into its own class (so that every class that schedules something will use the same scheduler).

That is mostly the client itself + this metrics reporter.

I think your concern is fair. I think in the latest PR is solved. Would you give it a look as well?

UnleashClient/async_metrics_reporter.py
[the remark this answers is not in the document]
ppicom: I had not thought of this 🤔 great point! Let me go over it one more time

UnleashClient/async_metrics_reporter.py
ppicom: that is a suuuuper valid point that I think I fixed in https://github.com/Unleash/unleash-python-sdk/pull/428, which is bad separation of PRs on my part... so apologies for that 🙇

UnleashClient/async_metrics_reporter.py
```diff
@@ -0,0 +1,92 @@
...
+        :param impact_metrics: drained for each send, and restored when a send fails.
+        """
+        self._config: UnleashConfig = config
+        self._transport: AsyncTransport = transport
+        self._scheduler: _AsyncScheduler = scheduler
+        self._engine: UnleashEngine = engine
+        self._impact_metrics: ImpactMetrics = impact_metrics
+        self._job: _AsyncScheduledJob = None
+
+    async def start(self) -> None:
```
alexanto: nothing is awaited here, why async?

alexanto: also, we could check here if there is already a job registered and not add a second one if called more than once

tests/unit_tests/clients/test_async_unleash_client.py
```diff
@@ -199,63 +201,6 @@ def test_both_clients_load_the_same_state(tmpdir):
         sync_client.destroy()
 
 
-def test_both_clients_build_the_same_kind_of_scheduler(tmpdir):
```
ppicom: these tests were testing internals that don't guarantee the integrity of our invariants, so I removed them.,

tests/unit_tests/clients/test_async_unleash_client.py
```diff
@@ -349,3 +294,53 @@ def test_the_async_client_gets_its_own_evaluator(tmpdir):
...
+def test_async_client_builds_a_metrics_reporter_over_its_own_collaborators(tmpdir):
+    client = build_async_client(tmpdir, url=URL, app_name=APP_NAME)
+
+    assert client._metrics._config is client._config
+    assert client._metrics._transport is client._transport
+    assert client._metrics._engine is client._engine
+    assert client._metrics._impact_metrics is client.impact_metrics
+
+
+def test_the_async_client_gets_the_async_reporter(tmpdir):
```
ppicom: this one seems to bring little value as well.

tests/unit_tests/test_async_metrics_reporter.py
```diff
@@ -0,0 +1,423 @@
...
+    reporter._config.app_name = "renamed"
+    reporter._engine.count_toggle(COUNTED_FLAG, True)
+    await reporter.flush()
+
+    assert metrics_body(server, 0)["appName"] == APP_NAME
+    assert metrics_body(server, 1)["appName"] == "renamed"
+
+
+@mark.asyncio
+async def test_the_flush_goes_through_the_async_transport(reporter):
```
ppicom: Now I have mixed feelings about this test as well.

UnleashClient/async_metrics_reporter.py
ppicom: I was leaving the door open to a fully async lifecycle, but it has stopped making sense the moment you pointed it out.

ppicom: fixed!

UnleashClient/async_metrics_reporter.py
ppicom: My rationale is that you want the loop (aka `_run`) to continue to work even if a job blows out. Buuut, when stopping the whole thing you want to find out that something went wrong.

I'm open to alternatives though 🤔

UnleashClient/async_metrics_reporter.py
```diff
@@ -23,9 +19,20 @@ class AsyncMetricsReporter:
     :meth:`start` and :meth:`stop` must be awaited from the event loop the client runs
     on, and the loop must stay open for as long as metrics are being reported.
 
-    The request is built from the :class:`~UnleashClient.config.UnleashConfig` at send
-    time, so reassigning a client's ``unleash_*`` attributes takes effect from the next
-    send.  ``metrics_interval`` and ``metrics_jitter`` are read once, by :meth:`start`.
+    Example::
```
alexanto: loving that there is example here 💯

UnleashClient/async_metrics_reporter.py
alexanto: there was a second comment on this wdyt about that?

UnleashClient/async_metrics_reporter.py
alexanto: ok, i agree with that principle, the loop swallows it while stop lets it thorugh, makes sense to me :+1:

alexanto:
There are still some unresolved comments, please comment/fix/resolve those as well :)

All in all it looks solid now, approving (really this time :sweat_smile: )

UnleashClient/async_metrics_reporter.py
ppicom: oh, that is a good one! it's showing up on linear but not on github 🤔

Yeh, I think it makes sense :yes:

alexanto (approved):

ppicom:
on it! thanks Alexandra

UnleashClient/async_metrics_reporter.py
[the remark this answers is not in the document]
ppicom: Done!

UnleashClient/async_metrics_reporter.py
ppicom: Now only one job is scheduled.

PR #425, opened 2026-08-21
EG-4662 (8.5): Introduce new `AsyncTransport`

ppicom:
This should be part of Stack #411, but I haven't managed to split that first PR. This is so the `AsyncTransport` and the `Evaluator` aren't part of the same huge PR.

- **chore: install aiohttp as an optional-dependency**
- **feat: Implement new `AsyncTransport`**
- **feat: wire AsyncUnleashClient with an AsyncTransport**

### DIFF
--- a/UnleashClient/async_transport.py
+++ b/UnleashClient/async_transport.py
@@ -1,0 +1,292 @@
+"""Asynchronous HTTP transport, for the async Unleash client."""
+
+import asyncio
+import json
+from typing import Any, Dict, Optional
+
+from multidict import CIMultiDict
+
+from UnleashClient.config import UnleashConfig
+from UnleashClient.constants import FEATURES_URL, METRICS_URL, REGISTER_URL
+from UnleashClient.headers import HeaderFactory
+from UnleashClient.transport import FetchResult, normalized_url
+from UnleashClient.utils import LOGGER
+
+try:
+    import aiohttp
+except ImportError as exc:  # pragma: no cover
+    raise ImportError(
+        "AsyncUnleashClient requires aiohttp, which is not installed. "
+        "Install it with: pip install UnleashClient[async]"
+    ) from exc
+
+
+# The statuses urllib3's Retry is given in Transport.fetch_features.
+RETRY_STATUSES = frozenset({500, 502, 504})
+
+# What a retry is worth attempting for. Deliberately narrower than
+# aiohttp.ClientError: InvalidUrlClientError and NonHttpUrlClientError also
+# subclass it, and a malformed url is not going to fix itself on attempt two.
+RETRYABLE_ERRORS = (aiohttp.ClientConnectionError, asyncio.TimeoutError)
+
+# The two aiohttp raises for a url it cannot use at all -- the analogue of the
+# requests quartet Transport.register re-raises. Both subclass ClientError, so
+# they have to be caught ahead of it. aiohttp has no InvalidHeader: an illegal
+# header surfaces as a bare ValueError, which is not a ClientError and so
+# escapes register() without a clause of its own.
+FATAL_URL_ERRORS = (aiohttp.InvalidURL, aiohttp.NonHttpUrlClientError)
+
+
+async def _log_resp_info(resp: "aiohttp.ClientResponse") -> None:
+    LOGGER.debug("HTTP status code: %s", resp.status)
+    LOGGER.debug("HTTP headers: %s", resp.headers)
+    LOGGER.debug("HTTP content: %s", await resp.text())
+
+
+class AsyncTransport:
+    """
+    The asyncio twin of :class:`UnleashClient.transport.Transport`, and the
+    async half of the one colored leaf in the SDK.
+
+    It matches ``Transport`` method for method and returns the same
+    :class:`~UnleashClient.transport.FetchResult`, so everything downstream of a
+    request -- :class:`~UnleashClient.store.FeatureStore`, the payload builders,
+    the :class:`~UnleashClient.headers.HeaderFactory` -- is shared rather than
+    duplicated. ``aclose()`` is the one method with no sync counterpart.
+
+    Like ``Transport``, it reads the config and asks the header factory on every
+    request rather than capturing either, because the ``unleash_*`` properties
+    that back them are public and writable.
+
+    Two things differ from the sync class, both forced by aiohttp:
+
+    - ``custom_options`` are ``requests`` keyword arguments and do not transfer.
+      ``verify``, ``cert`` and ``proxies`` all raise ``TypeError`` here; the
+      aiohttp spellings are ``ssl`` and ``proxy``.
+    - One pooled ``ClientSession`` is held rather than a session per request,
+      and it is built on first use, not in ``__init__`` -- aiohttp resolves the
+      running loop eagerly, and ``AsyncUnleashClient.__init__`` is synchronous.
+      That binds a transport to whichever loop first used it, and makes
+      ``aclose()`` an obligation on the caller.
+    """
+
+    def __init__(self, config: UnleashConfig, headers: HeaderFactory) -> None:
+        """
+        :param config: read for the url, timeouts, retries, project and custom
+                       options.
+        :param headers: builds the header set each request needs.
+        """
+        self._config: UnleashConfig = config
+        self._headers: HeaderFactory = headers
+        self._session: Optional["aiohttp.ClientSession"] = None
+
+    async def _get_session(self) -> "aiohttp.ClientSession":
+        if self._session is None or self._session.closed:
+            self._session = aiohttp.ClientSession()
+        return self._session
+
+    def _timeout(self) -> "aiohttp.ClientTimeout":
+        # sock_connect/sock_read rather than total, because requests' scalar
+        # timeout bounds each socket operation and not the whole request. Under
+        # `total` a large feature payload over a slow but healthy link would
+        # start failing at a request_timeout that works today.
+        seconds = self._config.request_timeout
+        return aiohttp.ClientTimeout(
+            total=None, sock_connect=seconds, sock_read=seconds
+        )
+
+    async def aclose(self) -> None:
+        """
+        Close the pooled session, if one was ever opened.
+
+        Nothing calls this yet -- ``AsyncUnleashClient.destroy()`` is still
+        unimplemented, and closing is its job when it lands. Until then every
+        caller that makes a request owns the close, or aiohttp logs an unclosed
+        session on garbage collection.
+        """
+        session, self._session = self._session, None
+        if session is not None and not session.closed:
+            await session.close()
+
+    # pylint: disable=broad-except
+    # TODO: narrow the except clause to the same set of errors the sync transport
+    # so TransportError becomes part of the API of transports.
+    async def fetch_features(self, etag: str = "") -> FetchResult:
+        """
+        Fetch feature state, sending ``If-None-Match`` when an etag is known.
+
+        Never raises. Any failure -- an unexpected status, a connection error, a
+        bad key in ``custom_options`` -- is logged and returned as an empty
+        result, so a polling job can keep running against a server that is down.
+
+        :param etag: the cached etag, or "" to fetch unconditionally.
+        """
+        config = self._config
+        try:
+            LOGGER.info("Getting feature flag.")
+
+            # A CIMultiDict rather than a dict: the uppercase pair below
+            # collides with the lowercase one HeaderFactory.base() already
+            # carries, and aiohttp would put *both* spellings on the wire where
+            # requests folds them into one. Updating a CIMultiDict replaces,
+            # which is the behaviour the sync transport has always had.
+            headers = CIMultiDict(self._headers.polling())
+
+            # The features endpoint has always been sent an uppercase copy of
+            # the two identification headers. It stays here rather than in
+            # polling() because it is a quirk of this one endpoint.
+            headers.update(
+                {
+                    "UNLEASH-APPNAME": config.app_name,
+                    "UNLEASH-INSTANCEID": config.instance_id,
+                }
+            )
+
+            if etag:
+                headers["If-None-Match"] = etag
+
+            base_url = normalized_url(config.url, FEATURES_URL)
+            base_params = {}
+
+            if config.project_name:
+                base_params = {"project": config.project_name}
+
+            session = await self._get_session()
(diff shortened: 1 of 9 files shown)


UnleashClient/transport.py
```diff
@@ -27,7 +27,7 @@ class FetchResult(NamedTuple):
     not_modified: bool = False
 
 
-def _normalized_url(url: str, path: str) -> str:
+def normalized_url(url: str, path: str) -> str:
```
sighphyre, line 30: Why this change? We're unprivating this beast? What for?

pyproject.toml
```diff
@@ -35,6 +35,12 @@ dependencies=[
     "launchdarkly-eventsource",
 ]
 
+[project.optional-dependencies]
+# Required by the in-progress AsyncUnleashClient only.  3.9 is the floor because
+# it introduced InvalidUrlClientError/NonHttpUrlClientError, which
+# AsyncTransport.register re-raises.
```
sighphyre, line 41: We are 100% going to forget this comment! And it's going to live here for the next half decade!

Also, not really sure it's helpful. Can we just yeet it?

tests/unit_tests/clients/test_async_unleash_client.py
```diff
@@ -246,3 +246,19 @@ def test_constructing_the_async_client_does_not_start_the_scheduler(tmpdir):
...
+    assert client._transport._headers is client._headers
+
+
+def test_constructing_the_async_client_opens_no_session(tmpdir):
+    # No event loop is running here, and none is needed: aiohttp resolves the
+    # loop when a ClientSession is built, so the transport has to defer that to
+    # the first request.
+    client = build_async_client(tmpdir, url=URL, app_name=APP_NAME)
+
+    assert client._transport._session is None
```
sighphyre, line 264: Is this important? What problem would this be surfacing if this ever fails?

tests/unit_tests/clients/test_async_unleash_client.py
```diff
@@ -246,3 +246,19 @@ def test_constructing_the_async_client_does_not_start_the_scheduler(tmpdir):
     client = build_async_client(tmpdir, url=URL, app_name=APP_NAME)
 
     assert not client._scheduler.scheduler.running
+
+
+def test_async_client_builds_a_transport_over_its_config_and_headers(tmpdir):
+    client = build_async_client(tmpdir, url=URL, app_name=APP_NAME)
+
+    assert client._transport._config is client._config
+    assert client._transport._headers is client._headers
```
sighphyre, line 255: This feels a little fragile. We're rooting around in this thing's internal state to see that it was assembled in the way we expect today. But that's an implementation detail, this test could become invalid and the system still holds together just fine

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+    assert [key for key in sent if key.lower() == "unleash-appname"] == [
+        "UNLEASH-APPNAME"
+    ]
+    assert [key for key in sent if key.lower() == "unleash-instanceid"] == [
+        "UNLEASH-INSTANCEID"
+    ]
+
+    received = server.calls("GET", FEATURES_PATH)[0].headers
+    assert received["unleash-appname"] == APP_NAME
+    assert received["unleash-instanceid"] == INSTANCE_ID
```
sighphyre, line 259: This an implementation detail, I don't think this needs a test

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+@mark.asyncio
+async def test_fetch_features_sends_each_identification_header_once(
+    server, transport, mocker
+):
+    # Read off the request rather than off the wire: aiohttp folds repeated
+    # spellings of a header itself, so a server cannot tell a transport that
+    # merged them from one that did not.
```
sighphyre, line 268: Correct. Because it's an implementation detail 😅

I think we can yeet this one too tbh

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+    # endpoint adds an uppercase one. requests folds the two into a single
+    # header; aiohttp would put both spellings on the wire unless the merge
+    # happens in a CIMultiDict first.
+    headers = spy.call_args.kwargs["headers"]
+    assert headers.getall("unleash-appname") == [APP_NAME]
+    assert headers.getall("unleash-instanceid") == [INSTANCE_ID]
+
+
+@mark.asyncio
+async def test_fetch_features_strips_a_trailing_slash_from_the_url(server, transport):
```
sighphyre, line 284: 👍 ❤️

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+
+    result = await transport.fetch_features()
+
+    assert len(server.calls("GET", FEATURES_PATH)) == 1
+    assert json.loads(result.raw_state)["version"] == 1
+    assert result.etag == ETAG_VALUE
+
+
+@mark.asyncio
+async def test_fetch_features_swallows_a_bad_custom_option(server, build_transport):
```
sighphyre, line 304: Is this good? Do we want this? If I'm reading this right, if you pipe in garbage your client just never works. Is that a test we want to enforce?

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+
+@mark.asyncio
+async def test_register_reraises_a_missing_scheme_as_a_value_error(transport):
+    transport._config.url = "thisisnotavalidurl"
+
+    # test_uc_with_invalid_url only sees the ValueError that requests'
+    # MissingSchema subclasses; aiohttp's InvalidURL subclasses it too, so the
+    # sync client's contract carries over.
+    with pytest.raises(ValueError):
+        await transport.register({})
```
sighphyre, line 385: This isn't sufficiently covered by the above test?

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+    transport._config.request_timeout = 7
+    await transport.send_metrics(MOCK_METRICS_REQUEST)
+
+    # unleash_request_timeout and friends are public setters, so an
+    # AsyncTransport built in __init__ has to pick up a change made after
+    # initialize_client().
+    timeout = spy.call_args.kwargs["timeout"]
+    # sock_connect/sock_read rather than total: requests' scalar timeout
+    # bounds each socket operation, not the whole request.
+    assert timeout == aiohttp.ClientTimeout(total=None, sock_connect=7, sock_read=7)
```
sighphyre, line 464: Why is this test important? What does this tell us if it fails? Do we intend for the config to be dynamically updatable?

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+@mark.asyncio
+async def test_every_endpoint_gets_its_headers_from_the_factory(
```
sighphyre, line 468: 👍 ❤️

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+@mark.asyncio
+async def test_aclose_is_a_no_op_before_any_request(build_transport):
+    transport = build_transport()
+
+    await transport.aclose()
+    await transport.aclose()
+
+    assert transport._session is None
```
sighphyre, line 531: But equally this test could pass but be wrong if close just nulls out the session, which I wouldn't be surprised if it did.

Also does this failing tell us much?

tests/unit_tests/test_async_transport.py
```diff
@@ -0,0 +1,551 @@
...
+    await transport.aclose()
+
+    assert first.closed
+    assert transport._session is None
+
+    # A closed session is not reused: aiohttp refuses to send on one.
+    await transport.send_metrics(MOCK_METRICS_REQUEST)
+
+    assert transport._session is not None
+    assert transport._session is not first
```
sighphyre, line 551: I have very limited knowledge of Python's async patterns and what's the ecosystem generally expects (and to be honest, I tend to disagree with a LOT of what modern Python programs do). Buuuut, if I close a thing, I expect that thing to be closed. Forever. I kinda half expect a closed thing to get pissy with me if I later start asking it to do things.

This test is enforcing... uh... necromancy I guess. Is that a thing we want to allow? Do we care?

tests/utilities/fake_unleash_server.py
```diff
@@ -0,0 +1,173 @@
+import logging
+from collections import deque
+from typing import Any, Deque, Dict, List, NamedTuple, Optional, Tuple
+
+from aiohttp import web
+from aiohttp.test_utils import TestServer
+from multidict import CIMultiDictProxy
```
sighphyre, line 7: Not sure what this is but it's not built in Python and I don't see it in the requirements. Guessing it comes in as a transitive dependency of aiohttp, could we make it an explicit test import for nerds like me who want to understand their direct dependencies?

UnleashClient/async_transport.py
```diff
@@ -0,0 +1,292 @@
...
+from UnleashClient.transport import FetchResult, normalized_url
+from UnleashClient.utils import LOGGER
+
+try:
+    import aiohttp
+except ImportError as exc:  # pragma: no cover
+    raise ImportError(
+        "AsyncUnleashClient requires aiohttp, which is not installed. "
+        "Install it with: pip install UnleashClient[async]"
+    ) from exc
```
sighphyre, line 21: Noice! This is a great way of handling the optional dependency!

UnleashClient/transport.py
ppicom: because the two transports use it. but maybe it makes more sense to have them both inside of `transport.py` 🤔

UnleashClient/async_transport.py
```diff
@@ -0,0 +1,292 @@
...
+
+        Never raises. Any failure -- an unexpected status, a connection error, a
+        bad key in ``custom_options`` -- is logged and returned as an empty
+        result, so a polling job can keep running against a server that is down.
+
+        :param etag: the cached etag, or "" to fetch unconditionally.
+        """
+        config = self._config
+        try:
+            LOGGER.info("Getting feature flag.")
```
sighphyre, line 126: Huh. See this is wrong in the old sync implementation as of 7 years ago 😅. Might as well make it correct here haha

```suggestion
            LOGGER.info("Getting feature flags.")
```

UnleashClient/async_transport.py
```diff
@@ -0,0 +1,292 @@
...
+
+            # A CIMultiDict rather than a dict: the uppercase pair below
+            # collides with the lowercase one HeaderFactory.base() already
+            # carries, and aiohttp would put *both* spellings on the wire where
+            # requests folds them into one. Updating a CIMultiDict replaces,
+            # which is the behaviour the sync transport has always had.
```
sighphyre, line 132: Okay entirely fair... sooo... why can't we just not do lines 138 through 143 then? If the headers we need are already set and the old implementation is doing something daft and setting them twice with different casing... could we just... not do that?

UnleashClient/async_transport.py
```diff
@@ -0,0 +1,292 @@
...
+    Two things differ from the sync class, both forced by aiohttp:
+
+    - ``custom_options`` are ``requests`` keyword arguments and do not transfer.
+      ``verify``, ``cert`` and ``proxies`` all raise ``TypeError`` here; the
+      aiohttp spellings are ``ssl`` and ``proxy``.
+    - One pooled ``ClientSession`` is held rather than a session per request,
+      and it is built on first use, not in ``__init__`` -- aiohttp resolves the
+      running loop eagerly, and ``AsyncUnleashClient.__init__`` is synchronous.
+      That binds a transport to whichever loop first used it, and makes
+      ``aclose()`` an obligation on the caller.
```
sighphyre, line 70: I think this is missing the hills for the trees. The differences are that the HTTP option have different names and that this class has to hand roll retries

sighphyre (approved):
I think the implementation is sound. Could use a human once over before merge because Claude is... *sigh* well.. Claude

tests/unit_tests/test_async_transport.py
ppicom: No, you're right. nulling out the session is not a guarantee of anything. An alternative could be to assert we don't throw.

But you're right. the test tells us very little, so I've deleted it 👍

tests/unit_tests/clients/test_async_unleash_client.py
ppicom: Yeah, no... I deffo tested too many internals.

tests/unit_tests/clients/test_async_unleash_client.py
ppicom: Uhm, yeah, I see what you're saying. Let me delete the test.

A question, though - would you test the collab between the client and the config somehow?

tests/unit_tests/test_async_transport.py
ppicom: Nope. In fact, we had an open conversation about what to do with custom options - the sync client let's our users "bypass" us and set `requests` options directly.

I believe we discussed that we might not want to let users do that for the async flavor (or any other flavor, for that matter)

So, I'm going to drop custom options altogether for the async client, and we can see what we want to do in a future release.

UnleashClient/async_transport.py
ppicom: made the comment simpler and more succint 👍

tests/unit_tests/test_async_transport.py
ppicom: I thought it would be, but it is true that that is a decision that comes with consequences 🤔 maybe we should not make it dynamically updatable: one "init client" == no changes after that.

tests/utilities/fake_unleash_server.py
ppicom: totally. listing it as a test dependency as we speak.

UnleashClient/async_transport.py
ppicom: yes, I've opted for... not doing that 🙌

tests/unit_tests/test_async_transport.py
ppicom: totally. much much better to just keep it closed. I've decided that it's a good idea to raise an error if you're using a closed client, so that as a user you're aware that you are using an SDK that doesn't necessarily work.

PR #428, opened 2026-09-23
EG-4662 (14) Implement `_AsyncScheduler` and wire it up in the `AsyncClient`

ppicom:
Continuing to build support for `async` in the python SDK, in this PR I introduce a new collaborator: `_AsyncScheduler` that will allow users to schedule a coroutine (aka an `async` function) as a recurring job.

Context: the SDK uses scheduling for things like reporting metrics at regular intervals.

There's one main principle driving the changes in these PRs: favor object composition over inheritance. In other words, the `UnleashClient` class was doing everything. So, to share the code, those "things" have been split up in different classes (that I often call collaborators because I'm like that). You may also read "leaf" (🍃 ) to refer to these collaborators: that is because there's two "trees" AsyncUnleashClient and UnleashClient sharing all the "leaves".

One thing to watch out, though: not all leaves can be shared. There is behavior that is straight up sync 100% of the time (like the `Evaluator` evaluating flags). We call these leaves "uncolored". There's other behavior that cannot be coded the same in sync code vs. async code.  We call these leaves "colored".

The `AsyncScheduler` is one of those things that has to be different in sync code vs. in async code.

- **feat: implement _AsyncScheduler**
- **feat: wire up _AsyncScheduler in the AsyncMetrics and the AsyncClient**


tests/unit_tests/clients/test_async_unleash_client.py
alexanto: should there not be a test that the async client gets the async scheduler now?

ppicom:
Closing this one, as it was folded into 11

PR #429, opened 2026-09-23
feat: implement _AsyncScheduler

ppicom:
this class is the collaborator that will enable `AsyncUnleashClient`
to run scheduler async functions as jobs.


UnleashClient/_async_scheduler.py
```diff
@@ -0,0 +1,166 @@
...
+        self._jobs: Set[_AsyncJob] = set()
+        self._running = False
+
+    def every(
+        self,
+        interval_seconds: float,
+        jitter_seconds: Optional[float],
+        fn: _AsyncJobFn,
+        kwargs: Optional[Dict[str, Any]] = None,
+    ) -> _AsyncScheduledJob:
```
alexanto: I don't see how this method can return None (maybe my shortcoming), but is the `_AsyncScheduledJob` really needed? cant the return type just be `_AsyncJob`

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 ## Unreleased
+
+> Note: this section of the changelog as-is it's working as a registry of all that
+> went on. It needs to be consolidated into what will eventually become the
+> final changelog + documentation chagnes (if any).
```
alexanto: just a typo :)
```suggestion
> final changelog + documentation changes (if any).
```

UnleashClient/_async_scheduler.py
```diff
@@ -0,0 +1,166 @@
...
+            if job.jitter:
+                delay += random.uniform(0, job.jitter)
+            await asyncio.sleep(max(0.0, delay))
+            try:
+                await job.fn(**job.kwargs)
+            except Exception:
+                LOGGER.exception("Scheduled job %r failed", job.fn)
+            next_at += job.interval
+            now = loop.time()
+            if next_at < now:
```
alexanto: this is part I commented about in the last PR, i see it skips missed  runs now, can we add a test for this?

tests/unit_tests/test_async_scheduler.py
```diff
@@ -0,0 +1,252 @@
...
+    assert len(runs) == runs_at_cancel
+
+
+@mark.asyncio
+async def test_cancel_tolerates_a_job_that_was_never_registered(scheduler):
+    scheduler.cancel(None)
+
+
+@mark.asyncio
+async def test_cancel_tolerates_a_job_that_is_already_gone(scheduler):
```
alexanto: no assertion in this test? 🤔

alexanto (commented):
nice PR, some small comments here and there, approving

tests/unit_tests/test_async_scheduler.py
ppicom: Uhm good one. The test is asserting that the thing doesn't panic on the second time. Buuut I can be more explicit about it with a try catch. Let me rework it.

UnleashClient/_async_scheduler.py
ppicom: Yes, you're right. that extra layer around the AsyncJob makes no sense.

ppicom: changed it so the API of the class is simpler. `every` gets you an `_AsyncJob`, you give `cancel` an `_AsyncJob`. If you pass `None` to these methods, then you're on your own.

UnleashClient/_async_scheduler.py
ppicom: Test added!

alexanto (approved):

